PR #17, opened 2022-10-03
feat: generate a lot of alert rules

gillesdemey:
This uses the provisioning HTTP API to generate a bunch of randomized alert rules with an interactive prompt.

image

### DIFF
--- a/scripts/generate-alert-rules/provision-large-group.mjs
+++ b/scripts/generate-alert-rules/provision-large-group.mjs
@@ -1,16 +1,20 @@
+#!/usr/bin/env node
+
 import axios from "axios"
 import pAll from "p-all"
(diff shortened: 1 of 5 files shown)


scripts/generate-alert-rules/provision-large-group.mjs
```diff
@@ -0,0 +1,109 @@
...
+          "type": "math",
+          "intervalMs": 1000,
+          "maxDataPoints": 43200
+        },
+        "datasourceUid": "-100",
+        "queryType": "",
+        "refId": "A"
+      }
+    ],
+    "for": "8h",
```
yuri-tceretian: I think this should be also configurable.

gillesdemey:
Closing this :)

PR #18, opened 2022-11-01
Refactor receivers to stop depending on `apimodels`

gotjosh:
To make sure we avoid circular dependencies, alerting needs to not depend on Grafana (it'll be the other way around - Grafana depends on Alerting).

This refactor removes the dependency from `apimodels` to do just that.

**This should break the tests and the linter due to a couple of missing functions but they would be immediately fixed on my next PR - I've chosen to split to make it easier to review**

### DIFF
--- a/alerting/receivers.go
+++ b/alerting/receivers.go
@@ -8,7 +8,6 @@ import (
 	"sort"
 	"time"
 
-	apimodels "github.com/grafana/grafana/pkg/services/ngalert/api/tooling/definitions"
 	"github.com/prometheus/alertmanager/notify"
 	"github.com/prometheus/alertmanager/types"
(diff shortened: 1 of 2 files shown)


alerting/receivers.go
```diff
@@ -186,9 +213,13 @@ func (am *GrafanaAlertmanager) TestReceivers(ctx context.Context, c apimodels.Te
 			return nil
 		})
 	}
-	g.Wait() // nolint
+	err = g.Wait() // nolint
```
gotjosh, line 216: The linter was complaining about this - I don't think the group can fail (because we handle failures) but worth checking to appease the linter.

grobinson-grafana (approved):

PR #19, opened 2022-11-02
Refactor GrafanaAlertmanager to not depend on Grafana

gotjosh:
This is the first substantial PR at decoupling the Alertmanager logic from the Grafana-specific logic - you can reference that by the removal of the Grafana-based imports from the code and the scope of an interface for that specific logic.

There's a LOT more work TODO (hence the number of comments in the code), but these should be followed up immediately and are very obvious that things don't work. For the ones that we think would be more long term and I'm happy to create an issue to track them.


alerting/grafana_alertmanager.go
```diff
@@ -135,7 +112,32 @@ type MaintenanceOptions interface {
...
+type Template = template.Template
+type InhibitRule = config.InhibitRule
+type MuteTimeInterval = config.MuteTimeInterval
+type Route = config.Route
+type Integration = notify.Integration
+type DispatcherLimits = dispatch.Limits
+type Notifier = notify.Notifier
+
+// Configuration is an interface for accessing Alertmanager configuration.
+type Configuration interface {
```
gotjosh, line 135: This is the biggest piece of this work - we've abstracted everything that could be Grafana-specific in an interface, there's probably more than will either go to Grafana or come back here but I've tried to do the minimum for now to get something done.

alerting/grafana_alertmanager.go
```diff
@@ -135,7 +112,32 @@ type MaintenanceOptions interface {
 	MaintenanceFunc() (int64, error)
 }
 
+type Template = template.Template
```
gotjosh, line 125: https://yourbasic.org/golang/type-alias/

alerting/alerts.go
```diff
@@ -21,11 +22,17 @@ var (
 	ErrGetAlertGroupsBadPayload = fmt.Errorf("unable to retrieve alerts groups")
 )
 
-func (am *GrafanaAlertmanager) GetAlerts(active, silenced, inhibited bool, filter []string, receivers string) (apimodels.GettableAlerts, error) {
+type GettableAlerts = amv2.GettableAlerts
```
gotjosh, line 25: There will be a better place to organise this, but for now, I want to make the diffs minimal until we know for sure that this works for Grafana. Then we can refactor away at our liking.

yuri-tceretian (approved):
LGTM

JohnnyQQQQ (approved):
LGTM, let's go!

JacobsonMT (approved):

PR #23, opened 2022-12-15
Shadow both `PostableAlerts` and `PostableAlert`

gotjosh:
We need them to do the conversion between the alert type in Grafana and the Alertmanager.


yuri-tceretian (approved):
LGTM

PR #27, opened 2022-12-23
Bump dependencies

santihernandezc:
Dependencies updated.
The functions `template.FromGlobs()` now takes a slice of strings instead of an arbitrary number of `string` as a variadic argument.


alexmobo (approved):
LGTM

PR #28, opened 2023-01-06
Support v3 APIs and basic auth in Kafka Contact Point

ghost:
What?
---
This commit does the following,
- Adds basic auth to the Kafka Contact Point.
- Adds support for v3 Kafka APIs

Basic Auth
---
When creating a Kafka Rest Proxy Contact Point on Grafana, users should now be able to use basic auth.
They can specific a username and can choose to either specify the password or let the password be read from a file on
the grafana server.
Like this,

Screenshot 2023-01-08 at 11 34 10 PM

Additionally, when the user chooses to supply the password as a file on the server, the password is re-read from the file on
an API failure. This is to allow the password to be changed without needing to restart the grafana server.

v3 Kafka APIs
----
The existing Kafka Contact Point only supported v2 APIs. While this was fine till now - newer systems are moving towards v3. Confluent is already only deploying v3 APIs for their Kafka Cloud Clusters.
This commit adds support to allow using v3 APIs to publish messages to kafka.

When creating the contact point, the user can now choose the API version they would like to use. By default, v2 will be used.

Screenshot 2023-01-08 at 11 41 19 PM

Testing
----
1. Added tests in kafka_tests.go
2. Manually tested this against my grafana deployment. Existing v2 APIs continued to work fine. Tested this using the "Test Contact Point" in the "Edit Contact Point" page.
3. Manually tested that I was able to messages to Kafka v3
4. [Done] ~TODO: Verify that the contact points created before this change continue to load fine after upgrading to this.~ Tested upgrade from 9.3.1 to main(v9.4.0-pre (d9be87c22d)) The previously provisioned kafka contact point was working as expected after the upgrade.

Things I need help with
-----
~1. How do I build and test this with Grafana?~
2. How do I know/specify which Grafana's release will this go into?

CHANGELOG
-----
- Kafka Contact Point now supports basic auth.
- Kafka Contact Point now supports v3 APIs.

References
----
1. [Kafka v2/v3 API comparison doc](https://docs.confluent.io/platform/current/kafka-rest/api.html)
2. Corresponding changes in Grafana UI: https://github.com/grafana/grafana/pull/61123

Credits
------
Something similar was done in this PR. Which I referenced: https://github.com/grafana/grafana/pull/41227

### DIFF
--- a/alerting/notifier/channels/kafka.go
+++ b/alerting/notifier/channels/kafka.go
@@ -63,14 +63,6 @@ type kafkaSettings struct {
 	KafkaClusterID string `json:"kafkaClusterId,omitempty" yaml:"kafkaClusterId,omitempty"`
 }
 
-// The user can either enter the password in the Grafana UI
-// or provide a file path from which to read the password.
-// These values represent the two options.
-const (
-	PasswordSourceFromInput = "password"
-	PasswordSourceFromFile  = "passwordFile"
-)
-
 // The user can choose which API version to use when sending
 // messages to Kafka. The default is v2.
 // Details on how these versions differ can be found here:
@@ -154,12 +146,14 @@ func (kn *KafkaNotifier) notifyWithAPIV2(ctx context.Context, as ...*types.Alert
 	tmpl, _ := TmplText(ctx, kn.tmpl, as, kn.log, &tmplErr)
 
 	topicURL := strings.TrimRight(kn.settings.Endpoint, "/") + "/topics/" + tmpl(kn.settings.Topic)
+	if tmplErr != nil {
+		kn.log.Warn("failed to template Kafka url", "error", tmplErr.Error())
+	}
 
 	body, err := kn.buildBody(ctx, tmpl, as...)
 	if err != nil {
 		return false, err
 	}
-
 	if tmplErr != nil {
 		kn.log.Warn("failed to template Kafka message", "error", tmplErr.Error())
 	}
@@ -177,7 +171,7 @@ func (kn *KafkaNotifier) notifyWithAPIV2(ctx context.Context, as ...*types.Alert
 	}
 
 	if err := kn.ns.SendWebhook(ctx, cmd); err != nil {
-		kn.log.Error("Failed to send notification to Kafka", "error", err, "body", cmd.Body)
+		kn.log.Error("Failed to send notification to Kafka", "error", err, "body", body)
 		return false, err
 	}
 	return true, nil
@@ -191,6 +185,9 @@ func (kn *KafkaNotifier) notifyWithAPIV3(ctx context.Context, as ...*types.Alert
 	// For v3 the Produce URL is like this,
 	// <Endpoint>/kafka/v3/clusters/<Cluster ID>/topics/<Topic Name>/records
 	topicURL := strings.TrimRight(kn.settings.Endpoint, "/") + "/kafka/v3/clusters/" + tmpl(kn.settings.KafkaClusterID) + "/topics/" + tmpl(kn.settings.Topic) + "/records"
+	if tmplErr != nil {
+		kn.log.Warn("failed to template Kafka url", "error", tmplErr.Error())
+	}
 
 	body, err := kn.buildV3Body(ctx, tmpl, as...)
 	if err != nil {
@@ -218,7 +215,7 @@ func (kn *KafkaNotifier) notifyWithAPIV3(ctx context.Context, as ...*types.Alert
 	// by setting “Transfer-Encoding: chunked” header.
 	// For as long as the connection is kept open, the server will keep accepting records.
 	if err := kn.ns.SendWebhook(ctx, cmd); err != nil {
-		kn.log.Error("Failed to send notification to Kafka", "error", err, "body", cmd.Body)
+		kn.log.Error("Failed to send notification to Kafka", "error", err, "body", body)
 		return false, err
 	}
 	return true, nil
--- a/alerting/notifier/channels/kafka.go
+++ b/alerting/notifier/channels/kafka.go
@@ -91,6 +91,7 @@ func buildKafkaSettings(fc FactoryConfig) (*kafkaSettings, error) {
 	if settings.Details == "" {
 		settings.Details = DefaultMessageEmbed
 	}
+	settings.Password = fc.DecryptFunc(context.Background(), fc.Config.SecureSettings, "password", settings.Password)
 
 	if settings.APIVersion == "" {
 		settings.APIVersion = APIVersionV2
--- a/alerting/notifier/channels/kafka.go
+++ b/alerting/notifier/channels/kafka.go
@@ -32,6 +32,11 @@ type kafkaRecord struct {
 	IncidentKey string         `json:"incident_key,omitempty"`
 }
 
+type kafkaV3Record struct {
+	Type string      `json:"type"`
+	Data kafkaRecord `json:"data"`
+}
+
 type kafkaContext struct {
 	Type   string `json:"type"`
 	Source string `json:"src"`
@@ -49,15 +54,35 @@ type KafkaNotifier struct {
 }
 
 type kafkaSettings struct {
-	Endpoint     string `json:"kafkaRestProxy,omitempty" yaml:"kafkaRestProxy,omitempty"`
-	Topic        string `json:"kafkaTopic,omitempty" yaml:"kafkaTopic,omitempty"`
-	Description  string `json:"description,omitempty" yaml:"description,omitempty"`
-	Details      string `json:"details,omitempty" yaml:"details,omitempty"`
-	BasicAuth    bool   `json:"basicAuth,omitempty" yaml:"basicAuth,omitempty"`
-	PassFilePath string `json:"passwordFilePath,omitempty" yaml:"passwordFilePath,omitempty"`
-	AuthUser     string `json:"user,omitempty" yaml:"user,omitempty"`
-	AuthPass     string `json:"password,omitempty" yaml:"password,omitempty"`
+	Endpoint       string `json:"kafkaRestProxy,omitempty" yaml:"kafkaRestProxy,omitempty"`
+	Topic          string `json:"kafkaTopic,omitempty" yaml:"kafkaTopic,omitempty"`
+	Description    string `json:"description,omitempty" yaml:"description,omitempty"`
+	Details        string `json:"details,omitempty" yaml:"details,omitempty"`
+	BasicAuth      bool   `json:"basicAuth,omitempty" yaml:"basicAuth,omitempty"`
+	BasicAuthUser  string `json:"basicAuthUser,omitempty" yaml:"user,omitempty"`
+	PasswordSource string `json:"passwordSource,omitempty" yaml:"passwordSource,omitempty"`
+	PassFilePath   string `json:"passwordFilePath,omitempty" yaml:"passwordFilePath,omitempty"`
+	AuthPass       string `json:"password,omitempty" yaml:"password,omitempty"`
+	APIVersion     string `json:"apiVersion,omitempty" yaml:"apiVersion,omitempty"`
+	KafkaClusterID string `json:"kafkaClusterId,omitempty" yaml:"kafkaClusterId,omitempty"`
 }
+
+// The user can either enter the password in the Grafana UI
+// or provide a file path from which to read the password.
+// These values represent the two options.
+const (
+	PasswordSourceFromInput = "password"
+	PasswordSourceFromFile  = "passwordFile"
+)
+
+// The user can choose which API version to use when sending
+// messages to Kafka. The default is v2.
+// Details on how these versions differ can be found here:
+// https://docs.confluent.io/platform/current/kafka-rest/api.html
+const (
+	APIVersionV2 = "v2"
+	APIVersionV3 = "v3"
+)
 
 func buildKafkaSettings(fc FactoryConfig) (*kafkaSettings, error) {
 	var settings kafkaSettings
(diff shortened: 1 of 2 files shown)


ghost:
Hey Guys!
Do I need a corresponding change in [available_channels.go](https://github.com/grafana/grafana/blob/main/pkg/services/ngalert/notifier/channels_config/available_channels.go) as well?

ghost:
Related issue: https://github.com/grafana/grafana/issues/41176

alexweav:
Hi, thanks very much for this contribution :wave:

At first glance, this seems like a reasonable change!

> Do I need a corresponding change in [available_channels.go](https://github.com/grafana/grafana/blob/main/pkg/services/ngalert/notifier/channels_config/available_channels.go) as well?

Yes, this will automatically create UI elements to match your new fields, so they will be visible and editable in the UI.

Regarding password in the DB vs. in a file, there are a few additional considerations to take into account:
- Grafana [encrypts](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-database-encryption/) secrets stored in the database.
- Network-based provisioning systems (Terraform, etc) do not have access to the disk of the instance, and will not be able to create a password file.
- Many users expect all of Grafana's data to be in the database, and many Grafana deployments have totally volatile filesystems.
- In Grafana's High-Availability mode, a password file would have to be synchronized among all instances, where a password in the database only has to be set once.

Additionally, some pointers that may be of help here:

> a password changes shouldn't require the user to restart the grafana server

We offer a way around this, sending a `POST` request to `/api/admin/provisioning/alerting/reload` should [reload and re-sync](https://grafana.com/docs/grafana/latest/developers/http_api/admin/#reload-provisioning-configurations) the provisioning files without a server restart.

> I would prefer to store the kafka password in a kubernates secret and mount it onto the Grafana pod as a file.

The provisioning files also allow environment variables to be referenced. A common pattern that we see is to mount the secret as an environment variable rather than a file, then [have the provisioning file point to it.](https://grafana.com/docs/grafana/latest/administration/provisioning/#using-environment-variables)

Ghazanfar-CFLT:
Thanks for the review @alexweav !
>Yes, this will automatically create UI elements to match your new fields, so they will be visible and editable in the UI.

Sounds good, I will raise another PR for this.

I think I agree with your points of needing to allow for a password.
Would it be okay if I allowed both options? Either supply the password or supply a path to a file containing the password?

>The provisioning files also allow environment variables to be referenced. A common pattern that we see is to mount the secret as an environment variable rather than a file, then [have the provisioning file point to it](https://grafana.com/docs/grafana/latest/administration/provisioning/#using-environment-variables)

The problem with this is that changes in the secret will not propagate to the container's ENV variable.
From the [Kubernetes doc](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables),
>Note: If a container already consumes a Secret in an environment variable, a Secret update will not be seen by the container unless it is restarted. There are third party solutions for triggering restarts when secrets change.

This is where I am coming from: My workplace wants us rotating passwords very frequently - every 15 mins or so. So, I would prefer the password be in a secret and mounted as a file. I would like to avoid frequent restarts of Grafana.

Given that there are uses for both the approaches, shall I go ahead with supporting them both please?

ghost:
Thanks for the review @alexweav !
>Yes, this will automatically create UI elements to match your new fields, so they will be visible and editable in the UI.

Sounds good, I will raise another PR for this.

I think I agree with your points of needing to allow for a password.
Would it be okay if I allowed both options? Either supply the password or supply a path to a file containing the password?

>The provisioning files also allow environment variables to be referenced. A common pattern that we see is to mount the secret as an environment variable rather than a file, then [have the provisioning file point to it](https://grafana.com/docs/grafana/latest/administration/provisioning/#using-environment-variables)

The problem with this is that changes in the secret will not propagate to the container's ENV variable.
From the [Kubernetes doc](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables),
>Note: If a container already consumes a Secret in an environment variable, a Secret update will not be seen by the container unless it is restarted. There are third party solutions for triggering restarts when secrets change.

This is where I am coming from: My workplace wants us rotating passwords very frequently - every 15 mins or so. So, I would prefer the password be in a secret and mounted as a file. I would like to avoid frequent restarts of Grafana.

Given that there are uses for both the approaches, shall I go ahead with supporting them both please?

ghost:
@alexweav Here are the corresponding changes for the UI: https://github.com/grafana/grafana/pull/61123

ghost:
Given https://github.com/grafana/grafana/pull/60655, I am assuming that the earliest release this change can only go into is 9.4. Is that correct?

ghost:
Added upgrade test results to the PR description.

alerting/notifier/channels/kafka.go
```diff
@@ -48,12 +54,36 @@ type KafkaNotifier struct {
...
-	Description string `json:"description,omitempty" yaml:"description,omitempty"`
-	Details     string `json:"details,omitempty" yaml:"details,omitempty"`
+	Endpoint       string `json:"kafkaRestProxy,omitempty" yaml:"kafkaRestProxy,omitempty"`
+	Topic          string `json:"kafkaTopic,omitempty" yaml:"kafkaTopic,omitempty"`
+	Description    string `json:"description,omitempty" yaml:"description,omitempty"`
+	Details        string `json:"details,omitempty" yaml:"details,omitempty"`
+	BasicAuth      bool   `json:"basicAuth,omitempty" yaml:"basicAuth,omitempty"`
+	BasicAuthUser  string `json:"basicAuthUser,omitempty" yaml:"user,omitempty"`
+	PasswordSource string `json:"passwordSource,omitempty" yaml:"passwordSource,omitempty"`
+	PassFilePath   string `json:"passwordFilePath,omitempty" yaml:"passwordFilePath,omitempty"`
```
yuri-tceretian: I do not think that we are ready to support this in Grafana notifiers, at least in the current implementation.
1. I think this creates quite a serious security vulnerability because there is nothing to stop a malicious user that has access to API to specify a path to SQLlite database, or grafana config file and send it in the Authz header back to malicious API endpoint.
2. None of Grafana notifiers support loading secrets from file. This might change when we start replacing Grafana implementation of notifiers with the Alertmanager ones, and we will have to figure out how to make it secure.

cc @gotjosh what do you think?

alerting/notifier/channels/kafka.go
```diff
@@ -48,12 +54,36 @@ type KafkaNotifier struct {
...
+	AuthPass       string `json:"password,omitempty" yaml:"password,omitempty"`
```
yuri-tceretian: Let's rename it to Password to match the name of the json field.

alerting/notifier/channels/kafka.go
```diff
@@ -48,12 +54,36 @@ type KafkaNotifier struct {
...
 type kafkaSettings struct {
-	Endpoint    string `json:"kafkaRestProxy,omitempty" yaml:"kafkaRestProxy,omitempty"`
-	Topic       string `json:"kafkaTopic,omitempty" yaml:"kafkaTopic,omitempty"`
-	Description string `json:"description,omitempty" yaml:"description,omitempty"`
-	Details     string `json:"details,omitempty" yaml:"details,omitempty"`
+	Endpoint       string `json:"kafkaRestProxy,omitempty" yaml:"kafkaRestProxy,omitempty"`
+	Topic          string `json:"kafkaTopic,omitempty" yaml:"kafkaTopic,omitempty"`
+	Description    string `json:"description,omitempty" yaml:"description,omitempty"`
+	Details        string `json:"details,omitempty" yaml:"details,omitempty"`
+	BasicAuth      bool   `json:"basicAuth,omitempty" yaml:"basicAuth,omitempty"`
```
yuri-tceretian: I think we do not need this field. I think validating that both user and password are specified would be enough

alerting/notifier/channels/kafka.go
```diff
@@ -130,11 +207,108 @@ func (kn *KafkaNotifier) Notify(ctx context.Context, as ...*types.Alert) (bool,
...
+
+// Use the v3 API to send the alert notification.
+func (kn *KafkaNotifier) notifyWithAPIV3(ctx context.Context, as ...*types.Alert) (bool, error) {
+	var tmplErr error
+	tmpl, _ := TmplText(ctx, kn.tmpl, as, kn.log, &tmplErr)
+
+	// For v3 the Produce URL is like this,
+	// <Endpoint>/kafka/v3/clusters/<Cluster ID>/topics/<Topic Name>/records
+	topicURL := strings.TrimRight(kn.settings.Endpoint, "/") + "/kafka/v3/clusters/" + tmpl(kn.settings.KafkaClusterID) + "/topics/" + tmpl(kn.settings.Topic) + "/records"
+
```
yuri-tceretian, line 192: can you add check here
```
if tmplErr != nil {
		kn.log.Warn("failed to template Kafka URL", "error", tmplErr.Error())
	}
```

yuri-tceretian (commented):
Generally, good change. Only serious concern is about loading secrets from file.

alerting/notifier/channels/kafka.go
gotjosh: Agreed with Yuri here, this does not belong within Grafana. We already have a way in Mimir to unsupport these fields in Cloud. 

So when we move over, we'll get it for free.

ghost: Got it! That's a valid concern.
I would also like to point out that it is current supported to supply Grafana's DB password from a file here: https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#file-provider

Would it help if I limit the size of the password on the file? To say 32 characters? 

The reasons I wanted to support this is so that when Grafana is deployed in a kubernetes environment, I would prefer that the password is stores in a kubernetes secret and mounted onto the grafana container. This also allows me to rotate the password frequently without restarting grafana each time.

yuri-tceretian: The difference between these two cases is that Grafana config can only be changed via updating a config file or via environment variable, and that requires access to the host. In the case of notification policies, this is a purely API call, which has much more relaxed constraints. Basically, any user with `Editor` role is able to create a contact point.

> This also allows me to rotate the password frequently without restarting grafana each time.

You can use an API call to rotate secrets instead. It can be a sidecar container that would do that.

alerting/notifier/channels/kafka.go
ghost: Sorry, there was a check here, I added additional code in the middle and it got moved down to line 235. I'll fix this!

alerting/notifier/channels/kafka.go
ghost: Would it help if I limit the size of the password on the file? To say 32 characters? Or would you guys prefer that I remove this altogether?

alerting/notifier/channels/kafka.go
ghost: Sounds good. I'll remove this.

alerting/notifier/channels/kafka.go
ghost: Agreed, I'll make the change!

alerting/notifier/channels/kafka.go
ghost: Thanks @yuri-tceretian 
I'll remove this part and leave you guys an update. Thanks for the review!

yuri-tceretian: One way I can recommend: introduce support for kafka in the alertmanager repository https://github.com/prometheus/alertmanager where you can introduce support for loading secrets from files. Then if it is accepted, you can configure Grafana to route all alerts to external Alertmanager instead of the internal one.

I believe opensource community would thank you for that!

ghost: Ah, that makes sense! Thanks @yuri-tceretian! I'll see if I can find the time to work on that.

alerting/notifier/channels/kafka.go
ghost: Done!

alerting/notifier/channels/kafka.go
ghost: I have removed validation for these altogether. Give both, give either, don't give anything, I just pass it along. Does that work @yuri-tceretian ? Or would you prefer that I validate that if one is provided, the other should be too?

alerting/notifier/channels/kafka.go
ghost: Done!

alerting/notifier/channels/kafka.go
```diff
@@ -48,12 +53,25 @@ type KafkaNotifier struct {
...
 type kafkaSettings struct {
-	Endpoint    string `json:"kafkaRestProxy,omitempty" yaml:"kafkaRestProxy,omitempty"`
-	Topic       string `json:"kafkaTopic,omitempty" yaml:"kafkaTopic,omitempty"`
-	Description string `json:"description,omitempty" yaml:"description,omitempty"`
-	Details     string `json:"details,omitempty" yaml:"details,omitempty"`
+	Endpoint       string `json:"kafkaRestProxy,omitempty" yaml:"kafkaRestProxy,omitempty"`
+	Topic          string `json:"kafkaTopic,omitempty" yaml:"kafkaTopic,omitempty"`
+	Description    string `json:"description,omitempty" yaml:"description,omitempty"`
+	Details        string `json:"details,omitempty" yaml:"details,omitempty"`
+	BasicAuthUser  string `json:"basicAuthUser,omitempty" yaml:"user,omitempty"`
```
yuri-tceretian: can it be Username and JSON\yaml `username`?

alerting/notifier/channels/kafka.go
ghost: Have remove the reading from file part. Will mark this resolved. I will get to the alertmanager changes in the future.

alerting/notifier/channels/kafka.go
ghost: On it!

ghost: Done!

alerting/notifier/channels/kafka.go
ghost: From here: https://stackoverflow.com/a/74658311/4042839
it looks like username and/or password can be empty. So, I don't think we should validate that if one is provided, the other should be too.

@yuri-tceretian does that answer it? Can I mark this as resolved?

yuri-tceretian: Although I think this is not a good practice, I am ok.

alerting/notifier/channels/kafka.go
```diff
@@ -73,6 +91,16 @@ func buildKafkaSettings(fc FactoryConfig) (*kafkaSettings, error) {
 	if settings.Details == "" {
 		settings.Details = DefaultMessageEmbed
 	}
+
```
yuri-tceretian, line 95: Password is supposed to be a secret. So please add 
```suggestion
	settings.Password = fc.DecryptFunc(context.Background(), fc.Config.SecureSettings, "password", settings.Password)
```

yuri-tceretian:
LGTM after the suggestion above is committed.

alerting/notifier/channels/kafka.go
ghost: @yuri-tceretian I can add this if you prefer :)
But won't that be adding limitations on top of features allowed by Basic Auth? If Basic Auth allows username or password to be empty, then we should allow it as well, right?

alerting/notifier/channels/kafka.go
ghost: Added!

yuri-tceretian (approved):
LGTM. Thanks!

ghost:
Thank you for reviewing @yuri-tceretian !
Am I good to merge this?

ghost:
Oh, I can't merge it! It says "Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests."

ghost:
Can someone also look at the corresponding UI changers on grafana please? https://github.com/grafana/grafana/pull/61123

PR #33, opened 2023-01-10
Alerting: Fix comma separated values in default template

grobinson-grafana:
This commit fixes a bug in the default template where values after the first value would not be comma separated. For example:

```
A=1234, B=5678, C=9
```

would have been:

```
A=1234, B=5678C=9
```


alexweav (approved):
LGTM

PR #34, opened 2023-01-10
Allow the BuildReceiverIntegrationsFunc to be injected as part of the config

gotjosh:
Also backports a very small change to skip the `NamespaceUIDLabel` when processing alerts.


yuri-tceretian (approved):
LGTM

PR #35, opened 2023-01-11
Alerting: Separate exprs in default template

grobinson-grafana:
This pull request changes the default template to show `Exprs` separate from `Value`, as `Exprs` can be quite verbose.

Before this change:

``` 
Value: A=1234 sum(rate(metric[5m])), B=5678 mean(A)
```

After this change:

```
Exprs:
 - A = sum(rate(metric(5m))
 - B = mean(A)
Value: A=1234, B=5678
```


alexmobo (approved):
LGTM

PR #40, opened 2023-01-13
Kafka Contact Point: Fix v3 API URL when used against a Kafka Rest Proxy

ghost:
Context
----

Support for using v3 APIs when producing to kafka was added in this PR: https://github.com/grafana/alerting/pull/28
However, I only tested this against a Confluent Kafka Cluster from [confluent.io](https://www.confluent.io/confluent-cloud/)
Turns out that the API paths are slightly different for the [Kafka Rest Proxy](https://github.com/confluentinc/kafka-rest) and the confluent kafka cluster. Thanks @yuri-tceretian for bringing this to my attention! 

This commit fixes the URL to work with the Kafka Rest Proxy. Confluent Kafka users can make use of this by add a "/kafka" suffix to the endpoint. We will add this as a hint on the UI.

Testing
---
1. Added a test and modified existing tests.
2. Verified that I am able to send a test notification to both Kafka Rest Proxy and Confluent Kafka Cluster.


yuri-tceretian:
Thanks for addressing this problem! Now we just need to add a hint to users to use `http://localhost:0000/kafka` for Endpoint url

ghost:
> Thanks for addressing this problem! Now we just need to add a hint to users to use `http://localhost:0000/kafka` for Endpoint url

Yep! On it!

ghost:
Updated the UI here: https://github.com/grafana/grafana/pull/61123/commits/412278f18069b469d9c1d0d259128d1273f2863d

yuri-tceretian (approved):

PR #41, opened 2023-01-13
Refactor notifier package

yuri-tceretian:
During moving the code from Grafana repository I moved all pieces with as little changes as possible persisting the history. Now the switch is done and it is time to take it apart.
This PR moves logger, imageStorage, templates, configs and clients (a sender that sends webhook and email)  to dedicated packages. 
All receiver configs are moved to the config package (similar to Alertamanager structure).

Also, it copies a factory from https://github.com/grafana/grafana/blob/f990be58cbfe8635d4970e97e4a4fb63a9e1997c/pkg/services/ngalert/notifier/channels_config/factory.go
which will let us move more code from Grafana to alerting

NOTE: the diff does not really make sense but the new dir structure is.


gotjosh:
Discussed this offline with Yuri. We're moving into a different more aggressive direction.

PR #49, opened 2023-01-31
Refactor alerting package

### LINKED ISSUE #46: Proposal: Package Structure
```
.
├── receivers/
│   ├── alertmanager/
│   │   ├── config.go
│   │   └── alertmanager.go
│   └── sensu/
│       ├── config.go
│       └── sensu.go
├── images/
│   └── images.go
├── logging/
│   └── log.go
└── crypto/
    └── crypto.go
```

yuri-tceretian:
1. All code moved from alerting sub-package to root. This removes the duplicated path component, e.g "github.com/grafana/alerting/alerting/models" turns to "github.com/grafana/alerting/models"
2. Refactored channels package.
  - Package renamed to `receivers`
  - All receivers are moved into their own sub-packages
  - Logging interface is moved to `logging` package in the root
  - Code for working with images is moved to `images` package in the root
  - Templates, structs and functions to work with them are moved to `templates` package in the root
3. All receivers renamed to Notifier. So reference to receivers will look like `email.Notifier` and constructor is `email.New`
4. Removed a factory function *Factory that wrapped New function with a `ReceiverInitError` and replaced by a generic function in the Factory.
5. Copied receivers factory from Grafana.

Fixes https://github.com/grafana/alerting/issues/46


images/images.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+type ImageStore interface {
+	GetImage(ctx context.Context, token string) (*Image, error)
+}
+
+type UnavailableImageStore struct{}
+
+// GetImage returns the image with the corresponding token, or ErrImageNotFound.
+func (u *UnavailableImageStore) GetImage(ctx context.Context, token string) (*Image, error) {
```
gotjosh:
```suggestion
func (u *UnavailableImageStore) GetImage(_ context.Context, _ string) (*Image, error) {
```

images/testing.go
```diff
@@ -9,12 +9,12 @@ import (
 	"time"
 )
 
-type fakeImageStore struct {
+type FakeImageStore struct {
 	Images []*Image
 }
 
 // getImage returns an image with the same token.
```
gotjosh:
```suggestion
// GetImage returns an image with the same token.
```

images/testing.go
```diff
@@ -25,8 +25,8 @@ func (f *fakeImageStore) GetImage(_ context.Context, token string) (*Image, erro
 
 // newFakeImageStore returns an image store with N test images.
```
gotjosh:
```suggestion
// NewFakeImageStore returns an image store with N test images.
```

images/utils.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/prometheus/alertmanager/types"
+	"github.com/prometheus/common/model"
+
+	"github.com/grafana/alerting/logging"
+	"github.com/grafana/alerting/models"
+)
+
+const (
+	// ImageStoreTimeout should be used by all callers for calles to `Images`
+	ImageStoreTimeout time.Duration = 500 * time.Millisecond
```
gotjosh:
```suggestion
	ImageStoreTimeout = 500 * time.Millisecond
```

gotjosh: You can omit the type.

images/utils.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+	return nil
+}
+
+// The path argument here comes from reading internal image storage, not User
+// input, so we ignore the security check here.
+//
+//nolint:gosec
+func OpenImage(path string) (io.ReadCloser, error) {
```
gotjosh, line 78:
```suggestion
// OpenImage returns an the io representation of an image from the given path.
func OpenImage(path string) (io.ReadCloser, error) {
```

notify/factory.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/grafana/alerting/receivers/teams"
+	"github.com/grafana/alerting/receivers/telegram"
+	"github.com/grafana/alerting/receivers/threema"
+	"github.com/grafana/alerting/receivers/victorops"
+	"github.com/grafana/alerting/receivers/webex"
+	"github.com/grafana/alerting/receivers/webhook"
+	"github.com/grafana/alerting/receivers/wecom"
+)
+
+var receiverFactories = map[string]func(receivers.FactoryConfig) (NotificationChannel, error){
```
gotjosh, line 31: `NotificationChannel` is a legacy Grafana concept, is it possible to get rid of it? I'm fine with this part of the separate PR as this PR is already kind of hard to review as it is right now.

notify/factory.go
```diff
@@ -0,0 +1,98 @@
...
+	"wecom":                   wrap(wecom.New),
+	"webex":                   wrap(webex.New),
+}
+
+type NotificationChannel interface {
+	notify.Notifier
+	notify.ResolvedSender
+}
+
+func Factory(receiverType string) (func(receivers.FactoryConfig) (NotificationChannel, error), bool) {
```
gotjosh, line 58: Do these functions warrant tests?

notify/factory.go
```diff
@@ -0,0 +1,98 @@
+package notify
```
gotjosh, line 1: I gotta admit, I'm not a fan of this pattern *at all*. I spent a big portion of my career writing this type of code in ruby and 10 years later I continue to find this hard to read/understand.

The layer of abstraction for the sake of saving an initialisation of an error seems too much.

That being said, I consider this a nit - if you all agree this is OK feel free to ignore my comment.

receivers/line/config.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/grafana/alerting/templates"
+)
+
+type Config struct {
+	Token       string `json:"token,omitempty" yaml:"token,omitempty"`
+	Title       string `json:"title,omitempty" yaml:"title,omitempty"`
+	Description string `json:"description,omitempty" yaml:"description,omitempty"`
+}
+
+func BuildConfig(fc receivers.FactoryConfig) (*Config, error) {
```
gotjosh: nit: This is validating the configuration but it's called `BuildConfig` in all receivers - this feels misleading, I understand that this "technically" returning a configuration because it's doing the marshalling but the common pattern I see in most cases is to simply execute the unmarshall call inline and call `Validate()` on top of the built object.

gotjosh (commented):
LGTM, but please look at my comments.

I was hoping this was done per package as opposed to in one go so that can we have isolated, bike shedable conversations (as this is a good time to bike shed).

yuri-tceretian:
> I was hoping this was done per package as opposed to in one go so that can we have isolated, bike shedable conversations (as this is a good time to bike shed).

I tried to do so by package but when I touched something it just affected all other files. So, I decided to do it in kinda one change. I did not touch `notify` package though, as I am not sure what the end structure will be because we still have some code in the Grafana repository.

notify/factory.go
yuri-tceretian: It's just a composition of two interfaces. I think we can but not in this PR. We need to move code that builds integrations from Grafana first.

notify/factory.go
yuri-tceretian: Not sure to which line of code you refer to. Do you mean `wrap` function? If yes, I am not a fan of it either, this was made as a workaround for providing a validation of the configuration prior to saving it to the database.
Currently, there is a similar factory method for each receiver and it looks exactly the same. I just made it DRY-er.

receivers/line/config.go
yuri-tceretian: I do not have a strong opinion on this name. BuildConfig and Validate have pros and cons. Ideally, that should be boiled down to custom marshaller implementation for each struct but for now, decryption functionality does not let us do that (I have a ddoc for that).

notify/factory.go
yuri-tceretian: It probably does but this code is mostly moved as is from Grafana where it did not have the tests either. I am not sure what we can test here. Probably that all receivers are covered. Each receiver is covered by its own tests.

receivers/alertmanager/alertmanager.go
```diff
@@ -10,10 +10,14 @@ import (
...
 	"github.com/prometheus/common/model"
+
+	"github.com/grafana/alerting/images"
+	"github.com/grafana/alerting/logging"
+	"github.com/grafana/alerting/receivers"
 )
 
-type AlertmanagerConfig struct {
-	*NotificationChannelConfig
+type Config struct {
```
alexweav, line 19: The package-per-receiver is an amazing change :+1:

receivers/alertmanager/alertmanager.go
```diff
@@ -77,23 +70,23 @@ func buildAlertmanagerNotifier(fc FactoryConfig) (*AlertmanagerNotifier, error)
 	}, nil
 }
 
-// AlertmanagerNotifier sends alert notifications to the alert manager
-type AlertmanagerNotifier struct {
-	*Base
-	images   ImageStore
+// Notifier sends alert notifications to the alert manager
+type Notifier struct {
```
alexweav, line 74: We called the package `receivers` but generalized all the notifier types `Notifier` - WDYT on aligning this? I.e. calling the package two layers up `notifiers` or calling these structs `Receiver`?

Do "notifier" and "receiver" always mean the same thing in our context?

yuri-tceretian: The reason I chose package `receivers` and struct name `Notifier` is that:
- we have `notify` package, and having `notifiers` would look awkward. If we want to rename to notifiers then probably we need to rename `notify` package.
- Notifier implements upstream's interface Notifier that has method Notify. Receiver in Alertmanager terminology is a set of Integrations, where each Integration wraps a Notifier.

receivers/line/config.go
yuri-tceretian: I renamed BuildConfig to ValidateConfig.

notify/factory.go
gotjosh: I'd test what happens when a receiver exists and when it does not.

notify/factory.go
gotjosh: > Not sure to which line of code you refer to

The whole design within this file. It's called `factory` so given I cannot comment on files I tend to typically pick the first line of the file when referring to the whole file.

notify/factory.go
```diff
@@ -0,0 +1,98 @@
...
+	Cfg    receivers.NotificationChannelConfig
+}
+
+func (e ReceiverInitError) Error() string {
+	name := ""
+	if e.Cfg.Name != "" {
+		name = fmt.Sprintf("%q ", e.Cfg.Name)
+	}
+
+	s := fmt.Sprintf("failed to validate receiver %sof type %q: %s", name, e.Cfg.Type, e.Reason)
```
gotjosh:
```suggestion
	s := fmt.Sprintf("failed to validate receiver %s of type %q: %s", name, e.Cfg.Type, e.Reason)
```

gotjosh (approved):
LGTM

notify/factory.go
yuri-tceretian: Ah, definitely, I think we should change it to the way it is implemented in Alertmanager!

PR #51, opened 2023-02-03
Fix formatting of the error in receveris factory

yuri-tceretian:
In #49 we accidentally broke the format of the errors, and it broke some tests in Grafana. This to return the previous format where the space is added if name is not empty (few line above)


alexweav (approved):
LGTM

PR #54, opened 2023-02-09
Use custom Number deserialization in pushover and webhook

yuri-tceretian:
The `json.Number` implementation fails if the field is an empty string. However, that may be the case when the configuration was created in Grafana 8 and wasn't changed since then.

This PR implements a custom OptionalNumber to support empty strings.

### DIFF
--- a/receivers/number.go
+++ b/receivers/number.go
@@ -5,13 +5,17 @@ import (
 	"strings"
 )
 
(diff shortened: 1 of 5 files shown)


JacobsonMT (approved):
LGTM, can you put the "why" for this directly in the code as a comment. Basically the first paragraph of your PR description will do.

PR #55, opened 2023-02-09
Backport: Use custom Number deserialization in pushover and webhook

yuri-tceretian:
This is a backport of https://github.com/grafana/alerting/pull/54 to 9.4.x branch

### DIFF
--- a/alerting/notifier/channels/pushover_test.go
+++ b/alerting/notifier/channels/pushover_test.go
@@ -182,6 +182,48 @@ func TestPushoverNotifier(t *testing.T) {
 			expMsgError: nil,
 		},
 		{
+			name: "Integer fields as empty strings",
+			settings: `{
+					"userKey": "<userKey>",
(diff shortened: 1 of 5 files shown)


alerting/notifier/channels/pushover_test.go
```diff
@@ -181,6 +181,48 @@ func TestPushoverNotifier(t *testing.T) {
 			},
 			expMsgError: nil,
 		},
+		{
+			name: "Integer fields as empty strings",
```
gotjosh, line 185: I don't see a test for string types, am I missing something?

gotjosh (approved):
LGTM

alerting/notifier/channels/pushover_test.go
yuri-tceretian: it's the test above. It tests more stuff but covers that case. I am working on refactoring all these tests and make it more clear.

PR #56, opened 2023-02-09
Fix reading `authorization_scheme` in webhook receiver

yuri-tceretian:
During refactoring in https://github.com/grafana/grafana/pull/55517 I missed mapping of field `authorization_scheme` to the resulting settings struct. This resulted in `authorization_credentials` always being sent as `Bearer`. In other  words, It broke only custom schemes (other than Bearer)


alexweav (approved):
LGTM

PR #60, opened 2023-02-14
Move integration building logic from Grafana

yuri-tceretian:
This PR splits the building of the integration from the alertmanager configuration.
- Interface `Configuration` is updated to require a method `Receivers` that returns a collection of `ApiReceiver`
- The Grafana Alertmanager is updated to accept a new struct IntegrationsBuilder. That struct encapsulates all services that are required to create an integration and build a `Notifier` from `GrafanaReceiver` configuration. The code is moved from Grafana with minimal changes.

Relates to PR https://github.com/grafana/grafana/pull/63340

### DIFF
--- a/notify/factory.go
+++ b/notify/factory.go
@@ -1,11 +1,15 @@
 package notify
 
 import (
+	"encoding/base64"
+	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/prometheus/alertmanager/notify"
 
+	"github.com/grafana/alerting/images"
+	"github.com/grafana/alerting/logging"
 	"github.com/grafana/alerting/receivers"
 	"github.com/grafana/alerting/receivers/alertmanager"
 	"github.com/grafana/alerting/receivers/dinding"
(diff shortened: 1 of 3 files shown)


alexmobo (approved):
LGTM

notify/factory.go
```diff
@@ -96,3 +100,89 @@ func (e ReceiverInitError) Error() string {
 }
 
 func (e ReceiverInitError) Unwrap() error { return e.Err }
+
+type IntegrationsBuilder struct {
```
gotjosh, line 104: Can you please explain what your plan is with this structure? I don't think this should be split into a separate struct to satisfy the interface and instead just remove the interface itself. 

We already have one level of indirection with the receiver Factory, and now we're introducing one more - I find this makes the code hard to follow and diverges with how this is done in two other places (upstream and Mimir) if possible, I'd like to have fewer differences not more.

gotjosh (changes requested):

notify/factory.go
yuri-tceretian: I use structure to pass multiple services through 3 functions of the builder. Currently, they are part of the Configuration interface, which is much harder to understand and read than this. The alternative would be to add them to the Alertmanager struct but I do not think this is a good solution because 1. those services (images service, notifications, logger factory, etc) are used by notifiers only, 2. the Alertmanager struct already contains many fields.

The `Factory` you mentioned is basically useless everywhere else because it requires data that cannot be provided by anything but the builder. So it can be made private and then simplified as it won't be used by anything else. 

If you have a better idea, I would appreciate it if you could explain it. The way it looks right now in Grafana with the cascade of closures is way less readable.

gotjosh: > The alternative would be to add them to the Alertmanager struct but I do not think this is a good solution because 1. those services (images service, notifications, logger factory, etc) are used by notifiers only, 2. the Alertmanager struct already contains many fields.

I'd rather see them there. I don't see how this is a bad idea and struct with many fields is a pretty common thing in Go e.g. https://github.com/grafana/mimir/blob/main/pkg/ingester/ingester.go#L219-L275

I'm a big fan of keeping indirection low - as it makes it easier to reason about the system as a whole.

yuri-tceretian:
Closing this PR as there is no agreement on a way forward, and will go without builder struct

gotjosh:
I don't understand why this PR is closed - Pull Requests are discussions, and this is exactly what we were having here. 

I apologise if I came across as abrasive with my words but that was not my intention here - code refactors are a bit subjective and what seems obvious to you might not be for me at first glance.

For the future, I wish to not be shut down the moment a conversation gets though and instead that we work together to reach a consensus or outcome.

yuri-tceretian:
I will go with a way that will not require this struct. Some kind of variation of what exists now. This PR is because I wanted to just move pieces from Grafana to cut the dependency earlier and then change it internally. So, I closed it in favor of several PRs #61 #62 and #63

PR #61, opened 2023-02-14
Refactor ValidateConfig of all receivers to accept only what is needed

yuri-tceretian:
Currently, every ValidateConfig accepts entire `receivers.FactoryConfig` and takes only 3 fields at best: raw settings, secure settings map, and decrypt function. 
This PR refactors those functions to accept only two parameters (sometimes one):
- raw settings
- a decrypt function `func(key string, fallback string) string` as you can see it does not accept the secure settings (and context). The reason is to hide the way secure settings are stored. 

Also, currently, some functions return `Config` struct and some pointer `*Config`. This PR refactors those that return pointer to return struct.

No public API of this package are changed.

### DIFF
--- a/receivers/alertmanager/config.go
+++ b/receivers/alertmanager/config.go
@@ -1,7 +1,6 @@
 package alertmanager
 
 import (
-	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -17,13 +16,13 @@ type Config struct {
 	Password string
 }
 
-func ValidateConfig(fc receivers.FactoryConfig) (Config, error) {
+func ValidateConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Config, error) {
 	var settings struct {
 		URL      receivers.CommaSeparatedStrings `json:"url,omitempty" yaml:"url,omitempty"`
 		User     string                          `json:"basicAuthUser,omitempty" yaml:"basicAuthUser,omitempty"`
 		Password string                          `json:"basicAuthPassword,omitempty" yaml:"basicAuthPassword,omitempty"`
 	}
-	err := json.Unmarshal(fc.Config.Settings, &settings)
+	err := json.Unmarshal(jsonData, &settings)
 	if err != nil {
 		return Config{}, fmt.Errorf("failed to unmarshal settings: %w", err)
 	}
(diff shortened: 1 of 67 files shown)


receivers/factory.go
```diff
@@ -42,6 +44,10 @@ type FactoryConfig struct {
 	Logger   logging.Logger
 }
 
+func (fc *FactoryConfig) Decrypt(key, fallback string) string {
+	return fc.DecryptFunc(context.Background(), fc.Config.SecureSettings, key, fallback)
+}
```
yuri-tceretian, line 49: This is a wrapper to hide secure settings and context from notifier's config builder

gotjosh: Why?

receivers/testing/testing.go
```diff
@@ -1,24 +1,11 @@
...
 )
 
-func NewFactoryConfigForValidateConfigTesting(t *testing.T, m *receivers.NotificationChannelConfig) (receivers.FactoryConfig, error) {
-	tmpl := templates.ForTests(t)
-	tmpl.ExternalURL = ParseURLUnsafe("http://localhost")
-	return receivers.NewFactoryConfig(m, nil, DecryptForTesting, tmpl, &images.UnavailableImageStore{}, func(ctx ...interface{}) logging.Logger {
-		return &logging.FakeLogger{}
-	}, "1.2.3")
-}
-
```
gotjosh, line 21: I'm assuming that all this setup (the templates, the logger, etc) was never used in the places where we setup this configuration?

receivers/factory.go
gotjosh: Figured it out - this is a nice change - not sure if that `context.Background()` might cause us more hard than good in the end but let's roll with it for now.

receivers/alertmanager/config.go
```diff
@@ -17,13 +16,13 @@ type Config struct {
 	Password string
 }
 
-func ValidateConfig(fc receivers.FactoryConfig) (Config, error) {
+func ValidateConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Config, error) {
```
gotjosh: Now that I fully have the context of this change - I'd say let's do one more refactor here and rename this method on all receivers from `ValidateConfig` to `NewConfig`.

`ValidateConfig` is wrong and deceiving, when you take some primitives and return a new struct this is way more than a validation.

gotjosh (changes requested):
LGTM but please look at my comment, I think we need one more quick change.

receivers/factory.go
yuri-tceretian: > Why?

To hide implementation details of secrets storage from config parser, and reduce the scope for functions NewConfig. The context will be enclosed upstream (in Grafana) where it makes sense because it provides a secret service that requires context.

receivers/testing/testing.go
yuri-tceretian: Pretty much, yes

gotjosh (approved):
LGTM

PR #62, opened 2023-02-14
Parse notifier configs without creation of notifiers

yuri-tceretian:
In Grafana Alertmanager configuration we have very complex API models that are usually a combination of upstream configurations and our custom fields. Those models are propagated everywhere in the code deep down to the notification-building logic where it is parsed and converted to `Notifier`. In order to validate the configuration in API handlers we have to access the notifiers factory and stub many fields because that factory requires many services.

For example,
https://github.com/grafana/grafana/blob/f066e8cdcd178b783d0c62959bea593d3c0c30b0/pkg/services/ngalert/api/tooling/definitions/provisioning_contactpoints.go#L105-L128

This PR creates an intermediate layer that parses, decrypts secrets and validates settings of all notifiers in a single receiver, and returns either a valid configuration or error. A new method `ValidateAPIReceiver` accepts only parameters that are required for performing this task. 

A new struct GrafanaReceiverTyped resembles the receiver configuration from the upstream 
https://github.com/prometheus/alertmanager/blob/fe1b045c00204dc47a0f63728b6162e9d5cf9b5e/config/config.go#L883-L899 However, due to additional information that Grafana notifier hold (uid, type, name) I had to add a generic wrapper `NotifierConfig`

This will help make validation code in API much clear and reduce the scope of raw settings.

Based on https://github.com/grafana/alerting/pull/61

The new config struct looks like the upstream Alertmanager's
```go
// GrafanaReceiverTyped represents a parsed and validated APIReceiver
type GrafanaReceiverTyped struct {
	Name                string
	AlertmanagerConfigs []*NotifierConfig[alertmanager.Config]
	DingdingConfigs     []*NotifierConfig[dinding.Config]
	DiscordConfigs      []*NotifierConfig[discord.Config]
	EmailConfigs        []*NotifierConfig[email.Config]
	GooglechatConfigs   []*NotifierConfig[googlechat.Config]
	KafkaConfigs        []*NotifierConfig[kafka.Config]
	LineConfigs         []*NotifierConfig[line.Config]
	OpsgenieConfigs     []*NotifierConfig[opsgenie.Config]
	PagerdutyConfigs    []*NotifierConfig[pagerduty.Config]
	PushoverConfigs     []*NotifierConfig[pushover.Config]
	SensugoConfigs      []*NotifierConfig[sensugo.Config]
	SlackConfigs        []*NotifierConfig[slack.Config]
	TeamsConfigs        []*NotifierConfig[teams.Config]
	TelegramConfigs     []*NotifierConfig[telegram.Config]
	ThreemaConfigs      []*NotifierConfig[threema.Config]
	VictoropsConfigs    []*NotifierConfig[victorops.Config]
	WebhookConfigs      []*NotifierConfig[webhook.Config]
	WecomConfigs        []*NotifierConfig[wecom.Config]
	WebexConfigs        []*NotifierConfig[webex.Config]
}
```


notify/factory.go
```diff
@@ -61,38 +60,24 @@ func Factory(receiverType string) (func(receivers.FactoryConfig) (NotificationCh
...
+				Err: err,
+				// TODO it will be removed in the next PR
+				Cfg: &GrafanaReceiver{
+					UID:                   fc.Config.UID,
+					Name:                  fc.Config.Name,
+					Type:                  fc.Config.Type,
+					DisableResolveMessage: fc.Config.DisableResolveMessage,
+					Settings:              fc.Config.Settings,
+					SecureSettings:        nil,
+				},
```
yuri-tceretian, line 78: temporary, this will go away in #63

receivers/factory.go
```diff
@@ -32,6 +32,14 @@ type NotificationChannelConfig struct {
 	SecureSettings        map[string][]byte `json:"secureSettings"`
 }
 
+// NotifierInfo contains metadata of the notifier. Name, UID, Type, etc
+type NotifierInfo struct {
+	UID                   string
+	Name                  string
+	Type                  string
+	DisableResolveMessage bool
+}
```
yuri-tceretian, line 41: This struct will replace NotificationChannelConfig in #63

notify/receivers.go
```diff
@@ -292,3 +316,189 @@ func ProcessNotifierError(config *GrafanaReceiver, err error) error {
...
+					Type:                  receiver.Type,
+					DisableResolveMessage: receiver.DisableResolveMessage,
+				},
+				Settings: settings,
+			}
+			f(r)
+			return nil
+		}
+	}
+}
```
yuri-tceretian, line 524: this is a utility func to propagate errors in the case of validation failure.

notify/receivers.go
```diff
@@ -292,3 +316,189 @@ func ProcessNotifierError(config *GrafanaReceiver, err error) error {
 
 	return err
 }
+
+// GrafanaReceiverTyped represents a parsed and validated APIReceiver
+type GrafanaReceiverTyped struct {
```
yuri-tceretian: I think it should be called GrafanaReceiver.. but that name is already taken. Probably it makes sense to rename that to GrafanaNotifierSettings and this to GrafanaReceiverConfiguration

notify/receivers.go
```diff
@@ -47,12 +71,12 @@ type InvalidReceiverError struct {
...
-	Name                  string                 `json:"name"`
-	Type                  string                 `json:"type"`
-	DisableResolveMessage bool                   `json:"disableResolveMessage"`
-	Settings              map[string]interface{} `json:"settings"`
-	SecureSettings        map[string]string      `json:"secureSettings"`
+	UID                   string            `json:"uid"`
+	Name                  string            `json:"name"`
+	Type                  string            `json:"type"`
+	DisableResolveMessage bool              `json:"disableResolveMessage"`
+	Settings              json.RawMessage   `json:"settings"`
```
yuri-tceretian, line 78: this will actually help move forward with refactoring in Grafana because we have to do conversions back and forth for that field.

receivers/factory.go
```diff
@@ -32,6 +32,14 @@ type NotificationChannelConfig struct {
 	SecureSettings        map[string][]byte `json:"secureSettings"`
 }
 
+// NotifierInfo contains metadata of the notifier. Name, UID, Type, etc
```
gotjosh:
```suggestion
// NotifierInfo contains the metadata of the notifier.
```

receivers/factory.go
```diff
@@ -32,6 +32,14 @@ type NotificationChannelConfig struct {
...
+type NotifierInfo struct {
```
gotjosh:
```suggestion
type Metadata struct {
```

gotjosh: Why not call it exactly that?

notify/receivers.go
```diff
@@ -292,3 +316,189 @@ func ProcessNotifierError(config *GrafanaReceiver, err error) error {
...
+						Err:      errors.New("failed to decode secure setting"),
+					}
+				}
+				secureSettings[k] = d
+			}
+		}
+
+		var decryptFn receivers.DecryptFunc = func(key string, fallback string) string {
+			return decrypt(ctx, secureSettings, key, fallback)
+		}
```
gotjosh: nit: 

I'd like to move away from these nested functions, they make this part of the code really hard to read. 

Why not just make this a separate function, even better why not embrace the pointer to structs here all of this a function that's part of `APIReceiver`. I don't really understand the relationship between these struct at all.

The code as written appears to take a functional approach - which even though is well intended I personally find it highly un-idiomatic.

For example, in this function the main actor is:

```golang
	for _, receiver := range api.Receivers {
		err := parseConfig(receiver)
		if err != nil {
			return GrafanaReceiverTyped{}, &ReceiverValidationError{
				Cfg: receiver,
				Err: err,
			}
		}
	}
```

Yet, I'm presented with `parseConfig` first which forces to me read this wall before I even begin to understand what I'm doing as part of this function. 

A good example of the use of inline functions is when you care about concurrency and sending that execution to a goroutine  - this is not any of that.

notify/receivers.go
```diff
@@ -292,3 +316,189 @@ func ProcessNotifierError(config *GrafanaReceiver, err error) error {
...
+
+		switch strings.ToLower(receiver.Type) {
+		case "prometheus-alertmanager":
+			return createReceiver[alertmanager.Config](receiver)(alertmanager.NewConfig(receiver.Settings, decryptFn))(func(f *NotifierConfig[alertmanager.Config]) {
```
gotjosh: I personally find this very hard to read - if all that we're saving is `err != nil` I'd like to move away from this model. 

Discussed this offline, agreed to move away from this model.

gotjosh (changes requested):
Overall LGTM and I find this is a net positive, but please see my comments.

notify/receivers.go
```diff
@@ -292,3 +316,222 @@ func ProcessNotifierError(config *GrafanaReceiver, err error) error {
...
+}
+
+// NotifierConfig represents parsed GrafanaReceiver.
+type NotifierConfig[T interface{}] struct {
+	receivers.Metadata
+	Settings T
+}
+
+// ValidateAPIReceiver parses, decrypts and validates the APIReceiver. GrafanaReceiverTyped that contains configurations of all notifiers configurations for this receiver
+func ValidateAPIReceiver(ctx context.Context, api *APIReceiver, decrypt receivers.GetDecryptedValueFn) (GrafanaReceiverTyped, error) {
```
gotjosh: This function is untested - can you please write some tests for it?

notify/receivers.go
```diff
@@ -292,3 +316,222 @@ func ProcessNotifierError(config *GrafanaReceiver, err error) error {
...
+			})
+		}
+		if err != nil {
+			return GrafanaReceiverTyped{}, &ReceiverValidationError{
+				Cfg: receiver,
+				Err: fmt.Errorf("failed to parse notifier %s (UID: %s): %w", receiver.Name, receiver.UID, err),
+			}
+		}
+	}
+	return result, nil
```
gotjosh: nit: I'd re-structure this function.

Generally, functions tend to do four main things: Collecting input, handling failure, doing work and delivering results.

I'd stick to that order when possible.

In addition to that, the name of `ValidateAPIReceiver` is highly misleading - we're way past the name of validation. I suggest `BuildReceiverIntegrations`.

gotjosh: With the above, my suggestion is:

```golang
var ErrorAPIReceiverValidation = func(receiver *GrafanaReceiver, err error) error {
	return &ReceiverValidationError{
		Cfg: receiver,
		Err: fmt.Errorf("failed to parse notifier %s (UID: %s): %w", receiver.Name, receiver.UID, err),
	}
}

// ValidateAPIReceiver parses, decrypts and validates the APIReceiver. GrafanaReceiverTyped that contains configurations of all notifiers configurations for this receiver
func ValidateAPIReceiver(ctx context.Context, api *APIReceiver, decrypt receivers.GetDecryptedValueFn) (GrafanaReceiverTyped, error) {
	result := GrafanaReceiverTyped{
		Name: api.Name,
	}

	for _, receiver := range api.Receivers {
		// First, decode the secure receiver integrations settings.
		secureSettings, err := decodeSecretsFromBase64(receiver.SecureSettings)
		if err != nil {
			return GrafanaReceiverTyped{}, ErrorAPIReceiverValidation(receiver, err)
		}

		// Then, create the decryption function that's common for all integrations of this receiver.
		decryptFn := func(key, fallback string) string {
			return decrypt(ctx, secureSettings, key, fallback)
		}

		// Finally, parse the receiver configuration.
		err = parseReceiver(&result, receiver, decryptFn)
		if err != nil {
			return GrafanaReceiverTyped{}, ErrorAPIReceiverValidation(receiver, err)
		}
	}

	return result, nil
}
```

notify/receivers.go
```diff
@@ -292,3 +316,222 @@ func ProcessNotifierError(config *GrafanaReceiver, err error) error {
...
+		}
+		result.WebexConfigs = append(result.WebexConfigs, newNotifierConfig(receiver, cfg))
+	default:
+		return fmt.Errorf("notifier %s is not supported", receiver.Type)
+	}
+	return nil
+}
+
+func decodeSecretsFromBase64(secrets map[string]string) (map[string][]byte, error) {
+	// secure settings are already encrypted at this point
```
gotjosh:
```suggestion
	// Secure settings are already encrypted at this point.
```

gotjosh: If secure settings are already encoded at this point - why do we need to have a `decrypt` function passed over? Should we call the function `noopDecryptFn` instead then?

notify/receivers.go
```diff
@@ -292,3 +316,222 @@ func ProcessNotifierError(config *GrafanaReceiver, err error) error {
...
+	Err error
+	Cfg *GrafanaReceiver
+}
+
+func (e ReceiverValidationError) Error() string {
+	name := ""
+	if e.Cfg.Name != "" {
+		name = fmt.Sprintf("%q ", e.Cfg.Name)
+	}
+	s := fmt.Sprintf("failed to validate receiver %sof type %q: %s", name, e.Cfg.Type, e.Err.Error())
```
gotjosh, line 536:
```suggestion
	s := fmt.Sprintf("failed to validate notifier %sof type %q: %s", name, e.Cfg.Type, e.Err.Error())
```

gotjosh: Other errors seem to use notifier instead?

receivers/factory.go
```diff
@@ -32,6 +32,14 @@ type NotificationChannelConfig struct {
 	SecureSettings        map[string][]byte `json:"secureSettings"`
 }
 
+// Metadata contains the metadata of the notifier
```
gotjosh:
```suggestion
// Metadata contains the metadata of the notifier.
```

notify/receivers.go
gotjosh: Remind me again how is this going to be used that will benefit from generics?

notify/receivers.go
yuri-tceretian: this is a copy from another place as is. I did not add anything here

https://github.com/grafana/grafana/blob/f066e8cdcd178b783d0c62959bea593d3c0c30b0/pkg/services/ngalert/notifier/alertmanager.go#L345-L346

notify/receivers.go
yuri-tceretian: I think that is expected because this error is returned by ValidateReceiver method... not a particular notifier.

notify/receivers.go
yuri-tceretian: That is a way to have Metadata for all notifier configs

notify/receivers.go
yuri-tceretian: This function is tricky to test thoroughly because it will mostly test receiver's NewConfig. Although I generally cover everything possible with tests, in this case I decided that it is not worth it because of the reason above. Anyway, if you think this does make sense, I will add tests.

yuri-tceretian: I implemented tests [85dbe42](https://github.com/grafana/alerting/pull/62/commits/85dbe4219572cac4b7a4ff0e09f48a29a837f669)

notify/receivers.go
```diff
@@ -292,3 +316,225 @@ func ProcessNotifierError(config *GrafanaReceiver, err error) error {
...
+	WebexConfigs        []*NotifierConfig[webex.Config]
+}
+
+// NotifierConfig represents parsed GrafanaReceiver.
+type NotifierConfig[T interface{}] struct {
+	receivers.Metadata
+	Settings T
+}
+
+// BuildReceiverConfiguration parses, decrypts and validates the APIReceiver. GrafanaReceiverConfig that contains configurations of all notifiers configurations for this receiver
```
gotjosh:
```suggestion
// BuildReceiverConfiguration parses, decrypts and validates the APIReceiver.
```

notify/receivers.go
```diff
@@ -292,3 +316,225 @@ func ProcessNotifierError(config *GrafanaReceiver, err error) error {
...
+			return GrafanaReceiverConfig{}, &ReceiverValidationError{
+				Cfg: receiver,
+				Err: fmt.Errorf("failed to parse notifier %s (UID: %s): %w", receiver.Name, receiver.UID, err),
+			}
+		}
+	}
+	return result, nil
+}
+
+// parseNotifier parses receivers and populates corresponding field in GrafanaReceiverConfig. Returns error if configuration cannot be parsed
```
gotjosh:
```suggestion
// parseNotifier parses receivers and populates the corresponding field in GrafanaReceiverConfig. Returns an error if the configuration cannot be parsed.
```

receivers/alertmanager/testing.go
```diff
@@ -0,0 +1,13 @@
+package alertmanager
+
+// FullValidConfigForTesting a string representation of JSON object that contains all fields supported by the notifier Config. Can be used without secrets
```
gotjosh:
```suggestion
// FullValidConfigForTesting is a string representation of JSON object that contains all fields supported by the notifier Config. It can be used without secrets.
```

gotjosh: The same applies to every other instance of this comment.

receivers/alertmanager/testing.go
```diff
@@ -0,0 +1,13 @@
...
+const FullValidConfigForTesting = `{
+	"url": "https://alertmanager-01.com",
+	"basicAuthUser": "grafana",
+	"basicAuthPassword": "admin"
+}`
+
+// FullValidSecretsForTesting is a string representation of JSON object that contains all fields that can be overridden from secrets
```
gotjosh, line 10:
```suggestion
// FullValidSecretsForTesting is a string representation of a JSON object that contains all fields that can be overridden from secrets.
```

gotjosh: The same applies to every other instance of this comment.

gotjosh (approved):
LGTM

PR #63, opened 2023-02-14
Replace notify.Factory with a build function that accepts typed configuration

yuri-tceretian:
Based on changes made in #61 and #62. 
This PR introduces a factory function `notify.BuildReceiverIntegrations` that accepts parsed receiver configuration and creates Integrations. 

Now, this looks pretty much closer to how upstream Alertmanager creates integrations.

Some notable changes:
- All notifier constructors are replaced to accept only arguments they need instead of FactoryConfig. 
- FactoryConfig and the structs that it uses are deleted. Function GetDecryptedValueFn moved to `notify` package where it is used only once, which will make it simpler to delete in future.
- Instead if `NotificationSender` there are two interfaces WebhookSender and EmailSender. In vanilla Alertmanager they are different for each notifier (actually, each notifier creates one by itself). Therefore, instead of accepting a single instance of senders `BuildReceiverIntegrations` now accepts factory functions, which can return an error. For example, lin in upstream Alertmanager https://github.com/prometheus/alertmanager/blob/41eb1213bb1c7ce0aa9e6464e297976d9c81cfe5/notify/webhook/webhook.go#L46


logging/log.go
```diff
@@ -1,6 +1,6 @@
 package logging
 
-type LoggerFactory func(ctx ...interface{}) Logger
+type LoggerFactory func(loggerName string, ctx ...interface{}) Logger
```
yuri-tceretian, line 3: this makes it clear that first element is always logger name whereas the remaining go to context (that's how it is implemented in Grafana)

notify/factory.go
```diff
@@ -27,57 +31,146 @@ import (
...
+		}
+		createIntegrationWithWebhook = func(idx int, cfg receivers.Metadata, f func(logger logging.Logger, w receivers.WebhookSender) notificationChannel) {
+			w, e := newWebhookSender(cfg)
+			if e != nil {
+				errors.Add(fmt.Errorf("unable to build webhook client for %s notifier %s (UID: %s): %w ", cfg.Type, cfg.Name, cfg.UID, e))
+				return
 			}
+			createIntegration(idx, cfg, func(logger logging.Logger) notificationChannel {
+				return f(logger, w)
+			})
```
yuri-tceretian: these to anonymous methods help reduce duplicated code below.

notify/receivers.go
```diff
@@ -347,8 +347,12 @@ type NotifierConfig[T interface{}] struct {
 	Settings T
 }
 
+// GetDecryptedValueFn is a function that returns the decrypted value of
+// the given key. If the key is not present, then it returns the fallback value.
+type GetDecryptedValueFn func(ctx context.Context, sjd map[string][]byte, key string, fallback string) string
+
```
yuri-tceretian, line 353: This function is moved from the receivers package because it is not used there anymore.

logging/log.go
santihernandezc: I like this change, but wouldn't it mean that we'll be using a different interface than the one defined in Grafana's `infra` package? In order to use the logger, we would have to add this string as the first item in the `ctx` slice.
I don't think it's a big deal, just want to make sure we're aware of this.

notify/factory.go
```diff
@@ -27,57 +31,146 @@ import (
...
+func BuildReceiverIntegrations(
+	receiver GrafanaReceiverConfig,
+	tmpl *template.Template,
+	img images.ImageStore, // Used by some receivers to include as part of the source
+	newLogger logging.LoggerFactory,
+	newWebhookSender func(n receivers.Metadata) (receivers.WebhookSender, error),
+	newEmailSender func(n receivers.Metadata) (receivers.EmailSender, error),
+	orgID int64,
+	version string,
+) ([]*Integration, error) {
```
santihernandezc:
```suggestion
```

notify/factory.go
```diff
@@ -27,57 +31,146 @@ import (
...
-					Name:                  fc.Config.Name,
-					Type:                  fc.Config.Type,
-					DisableResolveMessage: fc.Config.DisableResolveMessage,
-					Settings:              fc.Config.Settings,
-					SecureSettings:        nil,
-				},
+	var (
+		integrations      []*Integration
+		errors            types.MultiError
+		createIntegration = func(idx int, cfg receivers.Metadata, f func(logger logging.Logger) notificationChannel) {
```
santihernandezc: A comment explaining what this function does would be nice.

```suggestion
                // Helper function to create an integration for a notification channel and add it to the integrations slice.
		createIntegration = func(idx int, cfg receivers.Metadata, f func(logger logging.Logger) notificationChannel) {
```

notify/factory.go
```diff
@@ -27,57 +31,146 @@ import (
...
+			logger := newLogger("ngalert.notifier."+cfg.Type, "notifierUID", cfg.UID)
+			n := f(logger)
+			i := NewIntegration(n, n, cfg.Type, idx)
+			integrations = append(integrations, i)
+		}
+		createIntegrationWithWebhook = func(idx int, cfg receivers.Metadata, f func(logger logging.Logger, w receivers.WebhookSender) notificationChannel) {
```
santihernandezc: Same here, a comment would help.
```suggestion
                // Helper function to create an integration for a notification channel that requires a webhook sender.
		createIntegrationWithWebhook = func(idx int, cfg receivers.Metadata, f func(logger logging.Logger, w receivers.WebhookSender) notificationChannel) {
```

notify/factory.go
```diff
@@ -27,57 +31,146 @@ import (
...
+			if e != nil {
+				errors.Add(fmt.Errorf("unable to build webhook client for %s notifier %s (UID: %s): %w ", cfg.Type, cfg.Name, cfg.UID, e))
+				return
 			}
+			createIntegration(idx, cfg, func(logger logging.Logger) notificationChannel {
+				return f(logger, w)
+			})
+		}
+	)
+
```
santihernandezc:
```suggestion

// Range through each notification channel in the receiver and create an integration for it.
```

notify/factory.go
```diff
@@ -27,57 +31,146 @@ import (
...
-	"webhook":                 wrap(webhook.New),
-	"wecom":                   wrap(wecom.New),
-	"webex":                   wrap(webex.New),
-}
-
-type NotificationChannel interface {
-	notify.Notifier
-	notify.ResolvedSender
-}
+// BuildReceiverIntegrations builds notifiers of the receiver and wraps each of them in Integration.
```
santihernandezc: This comment could be more helpful.

```suggestion
// BuildReceiverIntegrations creates integrations for each configured notification channel in GrafanaReceiverConfig.
// It returns a slice of Integration objects, one for each notification channel, along with any errors that occurred.
```

santihernandezc (commented):
I like the idea of passing explicit values to the factory function instead of a `FactoryConfig` struct.
I requested a few comments to be added, but apart from that I think it's good to go!

logging/log.go
yuri-tceretian: Regularly, in the logging context is a slice of tuples. go-kit adds {Missing} if slice size is odd. However, In Grafana logger assumes that the first element is not a tuple's label but a logger name and automatically adds the `logger` label. However, this package can (and will) use a different logger implementation but usages of this interface still assume that the first element of the context is a logger. This change makes this interface more explicit.

> I like this change, but wouldn't it mean that we'll be using a different interface than the one defined in Grafana's infra package?

There is no interface for `New` method in infra. What I do here is basically define it. 

https://github.com/grafana/grafana/blob/862a6a2fa625100b690c5a48a9b12e7a7dd42041/pkg/infra/log/log.go#L231-L238

receivers/wecom/wecom.go
```diff
@@ -28,18 +28,14 @@ type Notifier struct {
...
-		Base:     receivers.NewBase(factoryConfig.Config),
-		tmpl:     factoryConfig.Template,
-		log:      factoryConfig.Logger,
-		ns:       factoryConfig.NotificationService,
-		settings: settings,
-	}, nil
+		Base:     receivers.NewBase(meta),
+		tmpl:     template,
+		log:      logger,
+		ns:       sender,
```
gotjosh, line 36: nit: This is called sender but the variable is installed called ns - this comment applies to all the others that use the webhook sender.

receivers/slack/slack.go
```diff
@@ -9,7 +9,7 @@ import (
 	"fmt"
 	"io"
 	"mime/multipart"
-	"net"
+	net "net"
```
gotjosh: nit: Why is this alias here?

receivers/discord/discord.go
```diff
@@ -82,20 +82,16 @@ type discordAttachment struct {
...
-		settings:   settings,
-		appVersion: fc.GrafanaBuildVersion,
-	}, nil
+		Base:       receivers.NewBase(meta),
+		log:        logger,
+		ns:         sender,
+		images:     images,
+		tmpl:       template,
+		settings:   cfg,
+		appVersion: buildVersion,
```
gotjosh:
```suggestion
		appVersion: appVersion,
```

receivers/alertmanager/alertmanager.go
```diff
@@ -13,18 +13,13 @@ import (
...
-	}
-
+func New(cfg Config, meta receivers.Metadata, images images.ImageStore, logger logging.Logger) *Notifier {
 	return &Notifier{
-		Base:     receivers.NewBase(fc.Config),
-		images:   fc.ImageStore,
-		settings: settings,
-		logger:   fc.Logger,
-	}, nil
+		Base:     receivers.NewBase(meta),
```
gotjosh, line 18: we filed #70 due to some problems we saw while reviewing this.

notify/factory.go
```diff
@@ -27,57 +31,147 @@ import (
...
-					DisableResolveMessage: fc.Config.DisableResolveMessage,
-					Settings:              fc.Config.Settings,
-					SecureSettings:        nil,
-				},
+// BuildReceiverIntegrations creates integrations for each configured notification channel in GrafanaReceiverConfig.
+// It returns a slice of Integration objects, one for each notification channel, along with any errors that occurred.
+func BuildReceiverIntegrations(
+	receiver GrafanaReceiverConfig,
+	tmpl *template.Template,
+	img images.ImageStore, // Used by some receivers to include as part of the source
```
gotjosh:
```suggestion
	img images.ImageStore,
```

gotjosh: This is pretty obvious - it causes more confusing by you trying to special-case it.

notify/factory.go
```diff
@@ -27,57 +31,147 @@ import (
...
+	newLogger logging.LoggerFactory,
```
gotjosh:
```suggestion
	logger logging.LoggerFactory,
```

notify/factory.go
```diff
@@ -27,57 +31,147 @@ import (
...
-					Type:                  fc.Config.Type,
-					DisableResolveMessage: fc.Config.DisableResolveMessage,
-					Settings:              fc.Config.Settings,
-					SecureSettings:        nil,
-				},
+// BuildReceiverIntegrations creates integrations for each configured notification channel in GrafanaReceiverConfig.
+// It returns a slice of Integration objects, one for each notification channel, along with any errors that occurred.
+func BuildReceiverIntegrations(
+	receiver GrafanaReceiverConfig,
+	tmpl *template.Template,
```
gotjosh, line 38:
```suggestion
	tmpl *alerting.Template,
```

you need the local types otherwise you'll import the alertmanage directly.

notify/factory.go
```diff
@@ -27,57 +31,147 @@ import (
...
+	img images.ImageStore, // Used by some receivers to include as part of the source
+	newLogger logging.LoggerFactory,
+	newWebhookSender func(n receivers.Metadata) (receivers.WebhookSender, error),
+	newEmailSender func(n receivers.Metadata) (receivers.EmailSender, error),
```
gotjosh, line 42: I think these functions should use primitives, such as:

```golang
func(ctx context.Context, client *http.Client, url string, body io.Reader) (*http.Response, error)
```

Instead of whatever interface was defined by Grafana with the `WebhookSender` and the `EmailSender`.

notify/factory.go
```diff
@@ -27,57 +31,147 @@ import (
...
+) ([]*Integration, error) {
+	type notificationChannel interface {
+		notify.Notifier
+		notify.ResolvedSender
+	}
+	var (
+		integrations []*Integration
+		errors       types.MultiError
+		// Helper function to create an integration for a notification channel and add it to the integrations slice.
+		createIntegration = func(idx int, cfg receivers.Metadata, f func(logger logging.Logger) notificationChannel) {
```
gotjosh: nit: If you're making me read this function from the get-go, I _think_ the name doesn't have to be verbose - I'd prefer a short name for the sake of clarity and readability

notify/factory.go
```diff
@@ -27,57 +31,147 @@ import (
...
+			logger := newLogger("ngalert.notifier."+cfg.Type, "notifierUID", cfg.UID)
+			n := f(logger)
+			i := NewIntegration(n, n, cfg.Type, idx)
+			integrations = append(integrations, i)
+		}
+		// Helper function to create an integration for a notification channel that requires a webhook sender.
+		createIntegrationWithWebhook = func(idx int, cfg receivers.Metadata, f func(logger logging.Logger, w receivers.WebhookSender) notificationChannel) {
```
gotjosh: Ditto, from here.

notify/factory.go
```diff
@@ -27,57 +31,147 @@ import (
...
+				return f(logger, w)
+			})
 		}
-		return ch, nil
+	)
+	// Range through each notification channel in the receiver and create an integration for it.
+	for i, cfg := range receiver.AlertmanagerConfigs {
+		createIntegration(i, cfg.Metadata, func(l logging.Logger) notificationChannel {
+			return alertmanager.New(cfg.Settings, cfg.Metadata, img, l)
+		})
```
gotjosh: nit: As we discussed offline there is another approach here to take, you can use more in-line function to try and simplify this createIntegration - it wasn't super clear why this was here.

```golang
		createIntegration(i, cfg.Metadata, dinding.New(cfg.Settings, cfg.Metadata, tmpl, nw(cfg.Metadata), nl(logger)))
```

gotjosh (approved):
LGTM

santihernandezc (approved):
All of my previous comments have been addressed 👍

receivers/wecom/wecom.go
yuri-tceretian: I will address it in a follow up PR as this rename will cause changes in other places.

notify/factory.go
yuri-tceretian: I will address it in a follow-up PR as it touches many lines.

PR #64, opened 2023-02-24
Metrics: Receivers and Integrations

gotjosh:
Introduce a set of metrics to observe the integrations and recivers of an instance.

For the sake of keeping cardinality as low as possible receviers are group into two categories: active or inactive However, for integrations I think it's useful to know how many we have of each type. I'd like to further know if an integration is active or not but decided not to do that know and roll with the current set in the meantime.

- [x] tests

### DIFF
--- a/notify/grafana_alertmanager_test.go
+++ b/notify/grafana_alertmanager_test.go
@@ -313,16 +313,6 @@ func TestPutAlert(t *testing.T) {
 	}
 }
 
-type fakeNotifier struct{}
-
-func (f *fakeNotifier) Notify(_ context.Context, _ ...*types.Alert) (bool, error) {
-	return true, nil
-}
-
-func (f *fakeNotifier) SendResolved() bool {
-	return true
-}
-
 func TestGrafanaAlertmanager_setReceiverMetrics(t *testing.T) {
 	fn := &fakeNotifier{}
 	integrations := []*notify.Integration{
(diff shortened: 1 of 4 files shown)


gotjosh:
Here's an example of how it looks like:

```
# HELP grafana_alerting_alertmanager_receivers Number of configured receivers by state. It is considered active if used within a route.
# TYPE grafana_alerting_alertmanager_receivers gauge
grafana_alerting_alertmanager_receivers{org="1",state="active"} 2
grafana_alerting_alertmanager_receivers{org="1",state="inactive"} 2
```

and

```
# HELP grafana_alerting_alertmanager_integrations Number of configured integrations within receivers by type.
# TYPE grafana_alerting_alertmanager_integrations gauge
grafana_alerting_alertmanager_integrations{org="1",type="discord"} 1
grafana_alerting_alertmanager_integrations{org="1",type="email"} 3
grafana_alerting_alertmanager_integrations{org="1",type="prometheus-alertmanager"} 1
grafana_alerting_alertmanager_integrations{org="1",type="slack"} 1
```

notify/grafana_alertmanager_test.go
```diff
@@ -309,12 +313,52 @@ func TestPutAlert(t *testing.T) {
...
 
+type fakeNotifier struct{}
+
+func (f *fakeNotifier) Notify(_ context.Context, _ ...*types.Alert) (bool, error) {
+	return true, nil
+}
+
+func (f *fakeNotifier) SendResolved() bool {
+	return true
+}
```
yuri-tceretian: I think this can live in receivers/testing

yuri-tceretian (approved):
LGTM

PR #65, opened 2023-03-14
Add CustomDetails field to PagerDuty

santihernandezc:
This PR enables users to add custom details to their PagerDuty integrations.

This field is an object containing arbitrary key-value pairs ([docs](https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTgw-events-api-v2-overview#pagerduty-common-event-format-pd-cef)). The field name is `custom_details`, as shown in the [example payload](https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTgx-send-an-alert-event#example-request-payloads) in PagerDuty's docs.

These details get merged with the default ones we're already sending. In case of a duplicate key, the user-defined value for that key prevails.

_Note: we should consider using a `map[string]interface{}` instead of `map[string]string` as we're currently doing. In the example I linked above, it's clear that the values in this object may not be always strings._


receivers/pagerduty/config_test.go
```diff
@@ -157,29 +157,33 @@ func TestNewConfig(t *testing.T) {
...
+					"num_resolved": "{{ .Alerts.Resolved | len }}",
+					"test":         "test",
+				},
+				Class:     DefaultClass,
+				Component: "Grafana",
+				Group:     DefaultGroup,
+				Summary:   templates.DefaultMessageTitleEmbed,
+				Source:    hostName,
+				Client:    DefaultClient,
+				ClientURL: "{{ .ExternalURL }}",
```
yuri-tceretian, line 253: Can you please add a test for overriding map values. Also, test for case sensitivity of keys

receivers/pagerduty/config.go
```diff
@@ -33,7 +44,7 @@ var getHostname = func() (string, error) {
 type Config struct {
 	Key           string            `json:"integrationKey,omitempty" yaml:"integrationKey,omitempty"`
 	Severity      string            `json:"severity,omitempty" yaml:"severity,omitempty"`
-	CustomDetails map[string]string `json:"-" yaml:"-"` // TODO support the settings in the config
+	CustomDetails map[string]string `json:"custom_details,omitempty" yaml:"custom_details,omitempty"`
```
yuri-tceretian: let's rename it to `Details` to match the upstream. https://github.com/prometheus/alertmanager/blob/41eb1213bb1c7ce0aa9e6464e297976d9c81cfe5/config/notifiers.go#L284

yuri-tceretian: including json field name

alexweav:
It seems there is also some handling of this field that prometheus does, and our notifier does not seem to do. Specifically, if the overall request is too big and will be rejected by PD, it makes sure to trim the custom details before anything else.

If we are going to make this field editable, let's consider handling it the same way?
https://github.com/prometheus/alertmanager/blob/41eb1213bb1c7ce0aa9e6464e297976d9c81cfe5/notify/pagerduty/pagerduty.go#L126-L142

receivers/pagerduty/config_test.go
santihernandezc: I updated a test to see how overriding keys would work.
https://github.com/grafana/alerting/blob/8882bca2cf0161fd1cd52acb2424bbd5cb3cabf8/receivers/pagerduty/config_test.go#L160-L169
The `"resolved"` key is part of our custom details, so when the final configuration has `"test"` as a value, it means that the default details were successfully overwritten.

I didn't consider adding case-sensitivity tests. I'll make a new commit with test cases for these two scenarios.

receivers/pagerduty/config.go
santihernandezc: D you think that's a good idea?
- In the docs I linked to in the description, the field is called "Custom Details". There's no mention of a "Details" field.
- The JSON tag for this field is also `"custom_details"`.
- Upstream converts this `"details"` field into `"custom_details"` when it has to send a notification, so it adds an extra step.

https://github.com/prometheus/alertmanager/blob/14a265b71024b858dd8186bd6ace3f666f1e934c/notify/pagerduty/pagerduty.go#L249

yuri-tceretian: Yes, our main target is that our config would be as close as possible to upstream Alertmanager config.

receivers/pagerduty/config_test.go
yuri-tceretian: I would prefer two tests one normal and one (or many) with overrides, so in the test name it would be crystal clear what it does

receivers/pagerduty/config.go
santihernandezc: That makes sense. I'll try to be as close to upstream as possible in my next commit.

receivers/pagerduty/pagerduty.go
```diff
@@ -74,15 +79,28 @@ func (pn *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error
...
 
-	body, err := json.Marshal(msg)
-	if err != nil {
-		return false, fmt.Errorf("marshal json: %w", err)
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(msg); err != nil {
+		return false, fmt.Errorf("failed to encode PagerDuty message: %w", err)
+	}
+
+	if buf.Len() > pagerDutyMaxEventSize {
```
yuri-tceretian, line 89: let's leave a comment with reference to alertmanager code base to reason about this solution

receivers/pagerduty/pagerduty.go
```diff
@@ -1,11 +1,14 @@
 package pagerduty
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
 	"strings"
 
+	"github.com/alecthomas/units"
```
yuri-tceretian, line 10: I am kinda on the fence about bringing this dependency for the sake of humanized bytes sizes... I think I am leaning towards keeping it because the upstream alertmaanger uses it also.

receivers/pagerduty/config_test.go
```diff
@@ -111,75 +111,146 @@ func TestNewConfig(t *testing.T) {
...
+				Group:     "test-group",
+				Summary:   "test-summary",
+				Source:    "test-source",
+				Client:    "test-client",
+				ClientURL: "test-client-url",
 			},
 		},
 		{
-			name: "Should ignore custom details",
+			name: "Shouldn't ignore details",
```
yuri-tceretian: Can we rename it to be something like "should merge with default details"

yuri-tceretian (approved):
LGTM! Great job

receivers/pagerduty/pagerduty.go
santihernandezc: I went through the same thought process, ultimately deciding on keeping it for the same reason.

PR #67, opened 2023-03-15
Slack: assume any non-JSON response is an incoming webhook

### LINKED ISSUE #64079: Slack notifications are repeated in Mattermost
**What happened**:
When using the Slack integration of Grafana Alerting in Mattermost, alerts are repeated according to the group interval, as Grafana incorrectly assumes that the notification was not sent successfully. Issue exists since upgrading Grafana from an older version to v9.4.2.
```
logger=ngalert.notifier.slack t=2023-03-02T19:45:11.745411891Z level=error msg="Failed to unmarshal response" body=ok err="invalid character 'o' looking for beginning of value"
logger=ngalert.notifier.slack t=2023-03-02T19:45:11.745663447Z level=error msg="Failed to send Slack message" err="failed to unmarshal response: invalid character 'o' looking for beginning of value"
logger=alertmanager org=1 t=2023-03-02T19:45:11.745821515Z level=error component=alertmanager orgID=1 component=dispatcher msg="Notify for alerts failed" num_alerts=2 err="Slack/slack[1]: notify retry canceled due to unrecoverable error after 1 attempts: failed to send Slack message: failed to unmarshal response: invalid character 'o' looking for beginning of value"
```
The following screenshot shows the error for a test notification:
Screenshot 2023-03-02 at 20 32 35

**Environment**:
- Grafana version: v9.4.2
- Mattermost version: v7.8.0

bfontaine:
Slack returns these responses as `text/html` while Mattermost returns them as `text/plain` (see https://github.com/grafana/grafana/issues/64079#issuecomment-1469826393). To support both, we need to either add another clause to the if (`|| strings.HasPrefix(content, "text/plain")`), either reverse the current logic, as does this PR.

I also added tests on the `Content-Type` header as they were missing. This fixes https://github.com/grafana/grafana/issues/64079.

### DIFF
--- a/receivers/slack/slack.go
+++ b/receivers/slack/slack.go
@@ -205,11 +205,11 @@ var sendSlackRequest = func(ctx context.Context, req *http.Request, logger loggi
 	}
 
 	content := resp.Header.Get("Content-Type")
-	// If the response is text/html it could be the response to an incoming webhook
-	if strings.HasPrefix(content, "text/html") {
-		return handleSlackIncomingWebhookResponse(resp, logger)
+	if strings.HasPrefix(content, "application/json") {
+		return handleSlackJSONResponse(resp, logger)
 	}
-	return handleSlackJSONResponse(resp, logger)
+	// If the response is not JSON it could be the response to an incoming webhook
+	return handleSlackIncomingWebhookResponse(resp, logger)
 }
 
 func handleSlackIncomingWebhookResponse(resp *http.Response, logger logging.Logger) (string, error) {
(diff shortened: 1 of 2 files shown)


leitmori:
@yuri-tceretian @grobinson-grafana Can you please take a look into this pull request to get notifications working again correctly in Mattermost?

yuri-tceretian (commented):
LGTM. Will leave to @grobinson-grafana to give a formal approval as we apply the practice of two reviewers per PR. Thanks for contribution!

yuri-tceretian:
I think this will need to be backported to 9.4.x branch after this one is merged. I can take care of it.

grobinson-grafana (approved):
Looks good! I also checked `chat.postMessage` and it returns `application/json` so this should work as expected!

PR #69, opened 2023-03-16
Alerting: Fix configuration of both incoming webhooks and bot tokens

### LINKED ISSUE #68: Slack allows configuration of both incoming webhooks and bot tokens at the same time
The Slack receiver allows configuration of both incoming webhooks and bot tokens at the same time. This is a bug. The intended behaviour is that users configure one or the other, but not both. The UI prevents users from being able to set both, but I think it's still possible when using the Provisioning API.

There are a number of incorrect tests that also assert both incoming webhooks and bot tokens can be configured at the same time. This is incorrect and should be fixed too.

grobinson-grafana:
This commit fixes a bug in Slack where it was possible to configure both incoming webhooks and bot tokens at the same time. This is a bug as the intended behaviour is that users configure one or the other, but not both.

Fixes #68 

Support escalation: https://github.com/grafana/support-escalations/issues/5309

### DIFF
--- a/receivers/slack/config_test.go
+++ b/receivers/slack/config_test.go
@@ -131,16 +131,15 @@ func TestNewConfig(t *testing.T) {
 		},
 		{
 			name:     "Should overwrite token from secrets",
-			settings: `{"url": "http://localhost", "token": "test" }`,
+			settings: `{ "recipient": "test-recipient", "token": "test" }`,
 			secureSettings: map[string][]byte{
-				"url":   []byte("http://slack.local/some-webhook"),
 				"token": []byte("test-token"),
 			},
 			expectedConfig: Config{
+				URL:            APIURL,
 				EndpointURL:    APIURL,
-				URL:            "http://slack.local/some-webhook",
 				Token:          "test-token",
-				Recipient:      "",
+				Recipient:      "test-recipient",
 				Text:           templates.DefaultMessageEmbed,
 				Title:          templates.DefaultMessageTitleEmbed,
 				Username:       "Grafana",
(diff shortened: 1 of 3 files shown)


receivers/slack/testing.go
```diff
@@ -3,7 +3,6 @@ package slack
 // FullValidConfigForTesting is a string representation of a JSON object that contains all fields supported by the notifier Config. It can be used without secrets.
 const FullValidConfigForTesting = `{
```
grobinson-grafana, line 4: I'm not sure what to do with these Yuri. Should I create one for incoming webhooks and another for bot tokens?

receivers/slack/config_test.go
```diff
@@ -256,7 +255,7 @@ func TestNewConfig(t *testing.T) {
 			settings: FullValidConfigForTesting,
 			expectedConfig: Config{
 				EndpointURL:    "http://localhost/endpoint_url",
-				URL:            "http://localhost/url",
+				URL:            "http://localhost/endpoint_url",
```
grobinson-grafana, line 258: This was changed to make the tests pass, but I think these tests need to be changed as they are based on the assumption you can configure both incoming webhooks and bot tokens at the same time.

receivers/slack/testing.go
yuri-tceretian: I would suggest to update this one and create another test with another combination of fields

JacobsonMT:
I'm not certain this is the way to go here, we have all the necessary information to send a notification correctly. Instead of erroring when both a URL and a token are provided, we could instead log a warning and use either the token or the webhook url (author's choice).

This is obviously niche since UI validation should prevent this in the first place, but if they do happen to get into this state (like in the escalation) maybe warning + continue as normal is better than error. What do you think?

grobinson-grafana:
> I'm not certain this is the way to go here, we have all the necessary information to send a notification correctly. Instead of erroring when both a URL and a token are provided, we could instead log a warning and use either the token or the webhook url (author's choice).

Once Slack is configured we don't know if its a token or a webhook because the URL field is re-used when using a webhook and the bot token. For example, when using an incoming webhook the URL is `https://hooks.slack.com/abc` and the token is `""`, but when using a bot token the URL is `https://api.slack.com` and the token is "abc".

PR #71, opened 2023-03-28
Use type alias for Template

yuri-tceretian:
This PR internalizes the usage of alertmanager's Template and updates all public functions and methods to use the type alias. This is a follow up for https://github.com/grafana/alerting/pull/63#discussion_r1146571465

### DIFF
--- a/receivers/googlechat/googlechat.go
+++ b/receivers/googlechat/googlechat.go
@@ -7,13 +7,12 @@ import (
 	"net/url"
 	"time"
 
-	"github.com/prometheus/alertmanager/template"
 	"github.com/prometheus/alertmanager/types"
 
 	"github.com/grafana/alerting/images"
 	"github.com/grafana/alerting/logging"
 	"github.com/grafana/alerting/receivers"
-	template2 "github.com/grafana/alerting/templates"
+	"github.com/grafana/alerting/templates"
 )
 
 // Notifier is responsible for sending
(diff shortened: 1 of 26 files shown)


gotjosh (approved):
LGTM

receivers/googlechat/googlechat.go
```diff
@@ -7,13 +7,12 @@ import (
 	"net/url"
 	"time"
 
-	"github.com/prometheus/alertmanager/template"
 	"github.com/prometheus/alertmanager/types"
```
santihernandezc, line 10: We're still importing the AM, right? related to [this comment](https://github.com/grafana/alerting/pull/63#discussion_r1146571465) on your other PR.

santihernandezc (commented):
I pulled your branch and saw we're still importing upstream's `template` package into ours.
- `default_template.go`: `Template` and `FromGlobs`are used in the `ForTests` function. Both of them are already part of our `template` package.
- `default_template_test.go`: This is a test file so I wouldn't worry about it.
- `funcs.go`: `DefaultFuncs` is taken from upstream.
- `template_data.go`:  There are some public types and functions (`ExtendData`, `TmplText`) that use the upstream package.


I don't think the imported types could be a concern, but the `ExtendData` and `TmplText` functions would force us to use the upstream package if we want to use them. Is this something that we could run into?

santihernandezc (approved):
LGTM 👍

PR #76, opened 2023-04-18
Return IntegrationValidationError instead of pointer

yuri-tceretian:
Changes error type that function notify.BuildReceiverConfiguration returns. The problem is that `errors.As(err, &IntegrationValidationError)` does not match the error (expected by Grafana code (https://github.com/grafana/grafana/blob/main/pkg/services/ngalert/api/api_alertmanager.go#L409)).


gotjosh (approved):
LGTM

PR #77, opened 2023-04-26
Add template testing functionality

JacobsonMT:
This adds template testing functionality.

**Description**
The method will attempt to interpolate each of the top-level definition blocks passed in through the template field with the given alerts as template context data. A top-level definition is one that isn't used by any others.

Default templates and any existing templates defined in the current AM will be available for reference by those being tested. The exception to this are the existing templates coming from the same filename as the one passed via the name field. This is so that users can real-time test out replacements to existing templates.

**Notes for reviewer**
There's some added complexity in this PR because template testing needs to recreate the existing templates from scratch. It needs to do this for a couple reasons:

1. Upstream Template does not provide a lot of way to manipulate its underlying `text/template` fields.
2. Even if it did, `text/template` does not provide a way to remove an already defined template. We need this in order to not mistakenly parse an existing template file that matches the name of the one we're testing as the new template might have removed some definitions. We could overwrite the templates with blanks but this might hide execution_errors.

So, in order to allow template testing to recreate the existing template we store the filenames and working directory for re-reading. I chose not to store the template text itself to keep the code as close as possible to existing template parsing which uses `templates.FromGlobs` instead of parsing from string.

Previously, the already parsed template was passed into `grafana_alertmanager` during `ApplyConfig`. This PR makes it so that `grafana_alertmanager` is the one that creates the template from the passed in filenames instead.


notify/receivers.go
```diff
@@ -246,6 +247,23 @@ func (am *GrafanaAlertmanager) TestReceivers(ctx context.Context, c TestReceiver
 	return newTestReceiversResult(testAlert, append(invalid, results...), now), nil
 }
 
+func (am *GrafanaAlertmanager) buildSingleIntegration(r *GrafanaIntegrationConfig, tmpl *templates.Template) (*Integration, error) {
```
JacobsonMT, line 250: This is [moved](https://github.com/grafana/grafana/blob/106eceab55627cb7d41acab2517f39a0d344e1bb/pkg/services/ngalert/notifier/alertmanager.go#L359-L374) from `grafana/grafana` which seems correct since it's used only for receiver testing.

notify/grafana_alertmanager.go
```diff
@@ -96,11 +97,15 @@ type GrafanaAlertmanager struct {
...
-	buildReceiverIntegrationFunc func(next *GrafanaIntegrationConfig, tmpl *templates.Template) (Notifier, error)
+	reloadConfigMtx sync.RWMutex
+	configHash      [16]byte
+	config          []byte
+	receivers       []*notify.Receiver
+
+	buildReceiverIntegrationsFunc func(next *APIReceiver, tmpl *templates.Template) ([]*Integration, error)
+	externalURL                   string
+	workingDirectory              string
+	templates                     []string
```
yuri-tceretian: please leave a few comments about what is `templates` and what is `template`

notify/grafana_alertmanager.go
```diff
@@ -140,18 +145,19 @@ type Configuration interface {
 	DispatcherLimits() DispatcherLimits
 	InhibitRules() []InhibitRule
 	MuteTimeIntervals() []MuteTimeInterval
-	ReceiverIntegrations() (map[string][]*Integration, error)
-	BuildReceiverIntegrationsFunc() func(next *GrafanaIntegrationConfig, tmpl *templates.Template) (Notifier, error)
+	Receivers() map[string]*APIReceiver
+	BuildReceiverIntegrationsFunc() func(next *APIReceiver, tmpl *templates.Template) ([]*Integration, error)
```
yuri-tceretian, line 156: NICE!

notify/receivers.go
yuri-tceretian: Testing API code needs to be refactored to build by receiver

notify/templates.go
```diff
@@ -0,0 +1,240 @@
...
+func findTopLevelTemplates(tmpl *tmpltext.Template) ([]string, error) {
+	// We need to find the names of all defined templates and subtract
+	// the names of all executed templates to find the set of templates that
+	// should be tested.
+	definedTmpls := make([]*tmpltext.Template, 0)
+	for _, def := range tmpl.Templates() {
+		// Check defined templates for an empty outer wrapper template.
+		// This can happen if the template filename does not match the name of any template definition.
+		// Remove if it exists.
+		if def.Name() == tmpl.ParseName && def.Root.Nodes == nil && def.Root.Pos == 0 {
```
grobinson-grafana: Can you explain the significance of `Nodes == nil` and `Pos == 0` as I'm not sure I understand why it's needed, but happy to have it explained to me.

I know that `len(Nodes) == 0` when the template contains just `{{ define "name" }}{{ end }}` blocks, but what about `Pos`?

notify/templates.go
```diff
@@ -0,0 +1,240 @@
...
+				Kind:    InvalidTemplate,
+				Message: err.Error(),
+			}},
+		}, nil
+	}
+
+	// Recreate the current template without the definition blocks that are being tested. This is so that any blocks that were removed don't get defined.
+	paths := make([]string, 0)
+	for _, name := range am.templates {
+		if name == c.Name {
```
grobinson-grafana, line 80: I don't think this works as expected? The following template called **other** doesn't return test2 or test3.

```
{{ define "test2" }}
test2
{{ end }}

{{ define "test3" }}
test3
{{ end }}
```

notify/receivers.go
```diff
@@ -246,6 +247,23 @@ func (am *GrafanaAlertmanager) TestReceivers(ctx context.Context, c TestReceiver
...
+			Integrations: []*GrafanaIntegrationConfig{r},
+		},
+	}
+	integrations, err := am.buildReceiverIntegrationsFunc(apiReceiver, tmpl)
+	if err != nil {
+		return nil, err
+	}
+	if len(integrations) == 0 {
+		// This should not happen, but it is better to return some error rather than having a panic.
+		return nil, fmt.Errorf("failed to build integration")
```
grobinson-grafana: I think you want to use `errors.New` here as otherwise the linter will complain that you have not passed any arguments to the format string?

```suggestion
		return nil, errors.New("failed to build integration")
```

notify/receivers.go
```diff
@@ -246,6 +247,23 @@ func (am *GrafanaAlertmanager) TestReceivers(ctx context.Context, c TestReceiver
...
+		GrafanaIntegrations: GrafanaIntegrations{
+			Integrations: []*GrafanaIntegrationConfig{r},
+		},
+	}
+	integrations, err := am.buildReceiverIntegrationsFunc(apiReceiver, tmpl)
+	if err != nil {
+		return nil, err
+	}
+	if len(integrations) == 0 {
+		// This should not happen, but it is better to return some error rather than having a panic.
```
grobinson-grafana, line 261: Can this happen?

notify/templates.go
```diff
@@ -0,0 +1,240 @@
...
+)
+
+type TestTemplatesConfigBodyParams struct {
+	// Alerts to use as data when testing the template.
+	Alerts []*PostableAlert `json:"alerts"`
+
+	// Template string to test.
+	Template string `json:"template"`
+
+	// Name of the template file.
```
grobinson-grafana, line 23: I was able to make it work without having this field required, but if we want to keep it we need to tell Sonia and Vika about it because it's not in the design doc.

notify/templates.go
```diff
@@ -0,0 +1,240 @@
...
+type TestTemplatesResult struct {
+	// Name of the associated template definition for this result.
+	Name string `json:"name"`
+
+	// Interpolated value of the template.
+	Text string `json:"text"`
+}
+
+type TestTemplatesErrorResult struct {
+	// Name of the associated template for this error. Will be empty if the Kind is "invalid_template".
```
grobinson-grafana, line 41: Good comments! :heart:

notify/templates.go
```diff
@@ -0,0 +1,240 @@
...
+
+	// Error message.
+	Message string `json:"message"`
+}
+
+type TemplateErrorKind string
+
+const (
+	InvalidTemplate TemplateErrorKind = "invalid_template"
+	ExecutionError  TemplateErrorKind = "execution_error"
```
grobinson-grafana, line 55: Nit: subsequent definitions should inherit the type from the above

```suggestion
	ExecutionError  = "execution_error"
```

notify/templates.go
```diff
@@ -0,0 +1,240 @@
...
+type TestTemplatesErrorResult struct {
+	// Name of the associated template for this error. Will be empty if the Kind is "invalid_template".
+	Name string `json:"name,omitempty"`
+
+	// Kind of template error that occurred.
+	Kind TemplateErrorKind `json:"kind"`
+
+	// Error message.
+	Message string `json:"message"`
+}
```
yuri-tceretian, line 49: json tags in these models make me think that they will be used for serialization, as API models. I think we should not have API models here and have an adapter layer in Grafana. WDYT?

notify/templates.go
```diff
@@ -0,0 +1,240 @@
...
+	ctx = notify.WithReceiverName(ctx, "test receiver")
+	ctx = notify.WithGroupLabels(ctx, model.LabelSet{"group_label": "group_label_value"})
+
+	var tmplErr error
+	templater, _ := templates.TmplText(ctx, newTmpl, alerts, am.logger, &tmplErr)
+
+	// Iterate over each definition in the template and evaluate it.
+	var results TestTemplatesResults
+	for _, def := range definitions {
+		s := fmt.Sprintf(`{{ template "%s" . }}`, def)
```
grobinson-grafana: I presume `"` is not allowed in template names, perhaps we should check just in case?

notify/templates.go
```diff
@@ -0,0 +1,240 @@
...
+// TestTemplate tests the given template string against the given alerts. Existing templates are used to provide context for the test.
+// If an existing template of the same filename as the one being tested is found, it will not be used as context.
+func (am *GrafanaAlertmanager) TestTemplate(ctx context.Context, c TestTemplatesConfigBodyParams) (*TestTemplatesResults, error) {
+	definitions, err := parseTestTemplate(c.Name, c.Template)
+	if err != nil {
+		return &TestTemplatesResults{
+			Errors: []TestTemplatesErrorResult{{
+				Kind:    InvalidTemplate,
+				Message: err.Error(),
+			}},
```
yuri-tceretian, line 73: I think we should return it as error and then convert to a response in API layer.

notify/templates.go
```diff
@@ -0,0 +1,240 @@
...
+	if err != nil {
+		// This shouldn't happen since we already parsed the template above.
+		return nil, err
+	}
+
+	externalURL, err := url.Parse(am.ExternalURL())
+	if err != nil {
+		return nil, err
+	}
+	newTmpl.ExternalURL = externalURL
```
yuri-tceretian: I think you can use TemplateFromPaths, modify it a little bit, so it will use externalURL from GrafanaALertmanager struct

notify/templates.go
```diff
@@ -0,0 +1,240 @@
...
+
+	// Prepare the context.
+	alerts := v2.OpenAPIAlertsToAlerts(c.Alerts)
+	ctx = notify.WithReceiverName(ctx, "test receiver")
+	ctx = notify.WithGroupLabels(ctx, model.LabelSet{"group_label": "group_label_value"})
```
yuri-tceretian: why do we need that? If we do need that then probably we should group c.Alerts and provide a correct label set?

notify/templates.go
```diff
@@ -0,0 +1,240 @@
...
+			Name: def,
+			Text: val,
+		})
+	}
+
+	return &results, nil
+}
+
+// parseTestTemplate parses the test template and returns the top-level definitions that should be interpolated as results.
+func parseTestTemplate(name string, template string) ([]string, error) {
```
yuri-tceretian: I think this and functions below should live in `templates` package because it's too specific for template processing

notify/templates.go
JacobsonMT: Basically, we're trying to detect the auto-created outer template that happens when the template filename doesn't match any of the given template definitions. If this happens it will be the first thing in the template (position 0), it will have no nodes other than other templates (nodes == nil) and the name of the template will be the same as the filename (`def.Name() == tmpl.ParseName`).

Consider this example, where the template filename is `slack.title`:
```
{{ define "slack.title" }}{{ template "default.title" . }}{{ end }}

{{- define "inner" -}}
{{- range .Labels.SortedPairs }}{{ .Name }}={{ .Value }} {{ end }} has value(s)
{{- range $k, $v := .Values }} {{ $k }}={{ $v }}{{ end }}
{{- end -}}

{{ define "outer" }}
There are {{ len .Alerts.Firing }} firing alert(s), and {{ len .Alerts.Resolved }} resolved alert(s)

{{ if .Alerts.Firing -}}
Firing alerts:
{{- range .Alerts.Firing }}
- {{ template "inner" . }}
{{- end }}
{{- end }}

{{ if .Alerts.Resolved -}}
Resolved alerts:
{{- range .Alerts.Resolved }}
- {{ template "inner" . }}
{{- end }}
{{- end }}
{{ end }}
```
I would expect a response like:
```
{
    "results": [
        {
            "name": "slack.title",
            "text": "..."
        },
        {
            "name": "outer",
            "text": "..."
        }
    ]
}
```

If, instead, the filename was something else, like `non_matching_filename`, IMO you should get the same response. If the above if statement didn't exist though, we would get an extra result or error with `name="non_matching_filename"`. It's arguable if we should hide this extra template or not, since it will technically exist when saved but it doesn't seem to add much value.

JacobsonMT: EDIT: I can't seem to reproduce this anymore, so I'm going to remove `pos==0` from the condition. Will leave this here for posterity.

If we left out `pos==0`, it wouldn't make that much difference but technically if they had an empty template defined with the same name as the file it would be left out accidentally, ex:

filename=`empty`
```
{{ define "slack.title" }}{{ template "default.title" . }}{{ end }}{{ define "empty" }}{{ end }}
```

Would return just `slack.title` in the response.

notify/templates.go
JacobsonMT: Just tested and it works for me. It did make me notice an issue with detecting the wrapper template though, that extra space in the middle means the wrapper has a node with empty space in it. I should check `parse.IsEmptyTree(def.Root)` instead of `def.Root==nil`.

notify/receivers.go
JacobsonMT: Unclear, I didn't write this function, just copied it from `grafana/grafana`.

notify/templates.go
JacobsonMT: If `def` contained a quote (or was otherwise invalid) it would not be parseable and would have failed with `invalid_template` earlier.

notify/templates.go
JacobsonMT: Good catch, yes I had originally copied this from the `grafana/grafana` api model. No need for them here.

notify/templates.go
JacobsonMT: Good idea :+1:

notify/templates.go
JacobsonMT: `GetTemplateData` will error if they don't exist. Otherwise, they are available as template values. We could group technically group them, but what would we use as the `groupby`? Do we actually use their notification policies or use a default of `folder, alertname`?

notify/templates.go
grobinson-grafana: I'll check again as I copied this comment from the original draft PR to avoid it getting lost 🙂

notify/templates.go
yuri-tceretian: Good question, I do not know. Probably as an alternative, we can group by common labels, or let user choose (in the API request)

notify/templates.go
JacobsonMT: This change gets caught by the linter as `SA9004: only the first constant in this group has an explicit type (staticcheck)`

notify/receivers.go
yuri-tceretian: It can't happen unless there is a bug in buildReceiverIntegrationsFunc that ignored one single integration specified in line 251

notify/templates.go
```diff
@@ -18,37 +17,37 @@ import (
 
 type TestTemplatesConfigBodyParams struct {
 	// Alerts to use as data when testing the template.
-	Alerts []*PostableAlert `json:"alerts"`
+	Alerts []*PostableAlert
```
yuri-tceretian, line 18: Ideally, I think we should accept []*types.Alert here because PostableAlert is API model in Alertmanager

notify/templates.go
JacobsonMT: This field was mandatory in the draft endpoint as well so I think we're ok for this.

notify/templates.go
JacobsonMT: Sounds good, I'll move `OpenAPIAlertsToAlerts` conversion to `grafana/grafana` side

notify/templates.go
JacobsonMT: Makes sense, I'll keep `parseTestTemplate` here but move `findTopLevelTemplates` and below to the templates package.

notify/templates.go
yuri-tceretian: please make an alias for it in alerting/notify package `OpenAPIAlertsToAlerts` so we do not make alertmanager explicit dependency.

templates/util_test.go
```diff
@@ -0,0 +1,68 @@
...
+	}, {
+		name:     "inline call to nested template that calls other nested template in if",
+		tmplName: "foo",
+		tmplText: `{{ define "baz" }}{{ range . }}The labels are {{.Labels }}{{ end }}{{ end }}{{ define "bar" }}{{ if len .Alerts }}{{ template "baz" .Alerts }}{{ end }}{{ end }}`,
+		expected: []string{"bar"},
+	}, {
+		name:     "inline call to nested template that calls other nested template in range",
+		tmplName: "foo",
+		tmplText: `{{ define "baz" }}The labels are {{.Labels }}{{ end }}{{ define "bar" }}{{ with .Alerts }}{{ template "baz" . }}{{ end }}{{ end }}`,
+		expected: []string{"bar"},
```
yuri-tceretian, line 56: I see the tests only assert a single item in the resulting slice. Can you add coverage for cases when there are multiple items?

notify/templates.go
```diff
@@ -0,0 +1,143 @@
...
+type TemplateErrorKind string
+
+const (
+	InvalidTemplate TemplateErrorKind = "invalid_template"
+	ExecutionError  TemplateErrorKind = "execution_error"
+)
+
+// TestTemplate tests the given template string against the given alerts. Existing templates are used to provide context for the test.
+// If an existing template of the same filename as the one being tested is found, it will not be used as context.
+func (am *GrafanaAlertmanager) TestTemplate(ctx context.Context, c TestTemplatesConfigBodyParams) (*TestTemplatesResults, error) {
```
yuri-tceretian, line 66: Do you plan to add tests for this functionality?

notify/templates.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/grafana/alerting/templates"
+	"github.com/prometheus/alertmanager/notify"
+	"github.com/prometheus/alertmanager/template"
+	"github.com/prometheus/alertmanager/types"
+	"github.com/prometheus/common/model"
+)
+
+type TestTemplatesConfigBodyParams struct {
+	// Alerts to use as data when testing the template.
+	Alerts []*types.Alert
```
yuri-tceretian: I think this will require alias for types.Alert too. Looked around in the code and we extensively use PostableAlerts in GrafanaAlertmanager API, so I am ok for returning it back even though I do not like that we use API models for internal communication.

notify/templates.go
JacobsonMT: I'm inclined to leave this as-is for now, it's not clear how to go about this. We can improve it in the next iteration.

notify/templates.go
grobinson-grafana: OK please just let Sonia and Vika know about it and also update the design doc 🙂

grobinson-grafana:
I get a recursive stack depth error for the following:

```
{
  "alerts": [{
    "annotations": {
      "summary": "Instance instance1 has been down for more than 5 minutes"
    },
    "labels": {
      "instance": "instance1"
    },
    "startsAt": "2023-04-01T00:00:00Z",
    "endsAt": "2023-04-01T00:05:00Z"
  }],
  "template": "{{ define \"test\" }}{{ range .Alerts }}{{ index .Annotations \"summary\" }}\n{{ end }}{{ end }}{{ template \"test\" . }}"
}
```

I think we should be able to make this work as per the following example:

```
package main

import (
	"bytes"
	"fmt"
	"text/template"
)

func main() {
	tmplText := `{{ define "hello" }}Hello, world!{{end}}{{ template "hello" . }}`
	tmpl, err := template.New("").Parse(tmplText)
	if err != nil {
		panic(err)
	}
	var buf bytes.Buffer
	tmpl.ExecuteTemplate(&buf, "", nil)
	fmt.Println(buf.String())
}
```

notify/templates.go
```diff
@@ -0,0 +1,149 @@
...
+	ctx = notify.WithReceiverName(ctx, DefaultReceiverName)
+	ctx = notify.WithGroupLabels(ctx, model.LabelSet{DefaultGroupLabel: DefaultGroupLabelValue})
+
+	var tmplErr error
+	templater, _ := templates.TmplText(ctx, newTmpl, alerts, am.logger, &tmplErr)
+
+	// Iterate over each definition in the template and evaluate it.
+	var results TestTemplatesResults
+	for _, def := range definitions {
+		s := fmt.Sprintf(`{{ template "%s" . }}`, def)
```
grobinson-grafana: Can we use `ExecuteTemplate` instead of creating a format string that we have to parse a second time? It should be faster and more robust, I shared an example in another comment.

notify/templates.go
grobinson-grafana: Still a bug for me 🙁 

```go
paths := make([]string, 0)
for _, name := range am.templates {
	fmt.Println("existing tmpl is", name)
	if name == c.Name {
		// Skip the existing template of the same name as we're going to parse the one for testing instead.
		continue
	}
	paths = append(paths, filepath.Join(am.workingDirectory, name))
}
```

```
existing tmpl is test1
existing tmpl is others
existing tmpl is __default__.tmpl
```

It doesn't print `test2` or `test3`:

Screenshot 2023-04-27 at 10 44 17

notify/templates.go
grobinson-grafana: I kind of understand, but I'm also kind of confused. 

If we take the example from [here](https://github.com/grobinson-grafana/template-test-example-1/blob/main/template_test.go#L19-L20) the template name is foo and it contains nodes for the `{{ range .Alerts }}{{ end }}`:

```go
tmplName: "foo",
tmplText: `{{ range .Alerts }}{{ end }}`,
```

If we do change tmplText to define foo again, the define in the tmplText just replaces the foo in `template.New()` such that the parsed template is the exact same as the example above:

```go
tmplName: "foo",
tmplText: `{{ define "foo" }}{{ range .Alerts }}{{ end }}{{ end }}`,
```

If we change tmplText to the following we have a template called *foo* that has no nodes because it just contains nested templates.

```go
tmplName: "foo",
tmplText: `{{ define "bar" }}{{ range .Alerts }}{{ end }}{{ end }}`,
```

And last if we do this we have a template called *foo* that contains a `NodeTemplate` node:

```go
tmplName: "foo",
tmplText: `{{ define "bar" }}{{ range .Alerts }}{{ end }}{{ end }}{{ template "bar" . }}`,
```

So I'm still a little confused about the need for `def.Name() == tmpl.ParseName && def.Root.Nodes == nil`. Which of the examples above is it supposed to work with?

notify/templates.go
```diff
@@ -0,0 +1,149 @@
...
+			// Skip the existing template of the same name as we're going to parse the one for testing instead.
+			continue
+		}
+		paths = append(paths, filepath.Join(am.workingDirectory, name))
+	}
+
+	// Parse current templates.
+	var newTextTmpl *tmpltext.Template
+	var captureTemplate template.Option = func(text *tmpltext.Template, _ *tmplhtml.Template) {
+		newTextTmpl = text
```
grobinson-grafana, line 90: What this does do? I tried printing the name of the template:

```go
fmt.Println("captured template is", text.Name())
```

but for the following example I just get a `""`:

```json
  "name": "test",
  "template": "{{ define \"test\" }}{{ range .Alerts }}{{ index .Annotations \"summary\" }}\n{{ end }}{{ end }}"
```

grobinson-grafana:
> I get a recursive stack depth error for the following:
> 
> ```
> {
>   "alerts": [{
>     "annotations": {
>       "summary": "Instance instance1 has been down for more than 5 minutes"
>     },
>     "labels": {
>       "instance": "instance1"
>     },
>     "startsAt": "2023-04-01T00:00:00Z",
>     "endsAt": "2023-04-01T00:05:00Z"
>   }],
>   "template": "{{ define \"test\" }}{{ range .Alerts }}{{ index .Annotations \"summary\" }}\n{{ end }}{{ end }}{{ template \"test\" . }}"
> }
> ```
> 
> I think we should be able to make this work as per the following example:
> 
> ```
> package main
> 
> import (
> 	"bytes"
> 	"fmt"
> 	"text/template"
> )
> 
> func main() {
> 	tmplText := `{{ define "hello" }}Hello, world!{{end}}{{ template "hello" . }}`
> 	tmpl, err := template.New("").Parse(tmplText)
> 	if err != nil {
> 		panic(err)
> 	}
> 	var buf bytes.Buffer
> 	tmpl.ExecuteTemplate(&buf, "", nil)
> 	fmt.Println(buf.String())
> }
> ```

This patch works for me :)

```
@@ -106,28 +106,26 @@ func (am *GrafanaAlertmanager) TestTemplate(ctx context.Context, c TestTemplates
 	ctx = notify.WithReceiverName(ctx, DefaultReceiverName)
 	ctx = notify.WithGroupLabels(ctx, model.LabelSet{DefaultGroupLabel: DefaultGroupLabelValue})

-	var tmplErr error
-	templater, _ := templates.TmplText(ctx, newTmpl, alerts, am.logger, &tmplErr)
+	promTmplData := notify.GetTemplateData(ctx, newTmpl, alerts, am.logger)
+	data := templates.ExtendData(promTmplData, am.logger)

 	// Iterate over each definition in the template and evaluate it.
 	var results TestTemplatesResults
 	for _, def := range definitions {
-		s := fmt.Sprintf(`{{ template "%s" . }}`, def)
-		val := templater(s)
-		if tmplErr != nil {
+		var buf bytes.Buffer
+		err := tmpl.ExecuteTemplate(&buf, def, data)
+		if err != nil {
 			results.Errors = append(results.Errors, TestTemplatesErrorResult{
 				Name:  def,
 				Kind:  ExecutionError,
-				Error: tmplErr,
+				Error: err,
+			})
+		} else {
+			results.Results = append(results.Results, TestTemplatesResult{
+				Name: def,
+				Text: buf.String(),
 			})
-			tmplErr = nil
-			continue
 		}
-
-		results.Results = append(results.Results, TestTemplatesResult{
-			Name: def,
-			Text: val,
-		})
 	}

 	return &results, nil
```

I'm not sure what you want to do with this code though as it's not required in the design doc for what we're going to deliver before Grafana 10.

```go
newTmpl, err := am.TemplateFromPaths(paths, captureTemplate)
if err != nil {
	return nil, err
}
```

notify/templates.go
grobinson-grafana: On further thought I think I want to disagree with adding `name`, or at least making it a required field.

The reason the design doc doesn't have a `name` field in the request JSON is because from a UX perspective I don't want users to have to choose a name in the UI as a pre-requisite to previewing templates. I don't think it's required for it to work and so all it does is create extra work for the user just to get a simple preview.

If and when the user is satisfied with the preview can they choose a name in the UI and click the Save button.

notify/templates.go
JacobsonMT: This one:

```
tmplName: "foo",
tmplText: `{{ define "bar" }}{{ range .Alerts }}{{ end }}{{ end }}`,
```

ParseName will be "foo" and when we iterate over the defined templates, one named "foo" will show up with empty nodes. We can either return that in the response with empty text or omit it from the response. If we omit it, we'll need that conditional.

notify/templates.go
JacobsonMT: In the screenshot you're printing the filename of existing persisted templates, not the name of the various definitions. Also, it won't know about any of the templates currently being tested, those aren't persisted.

notify/templates.go
JacobsonMT: Ah, I see what you're saying. I agree:

If a user wants to their test to be tested as if it replaces the currently existing template of the same name, it is required. Otherwise, if they're creating a new one and it's blank it will work fine without it in the current code.

notify/templates.go
JacobsonMT: The idea was to gain access to the underyling `tmpltest.Template` in order to parse in the new template to be tested. However, it looks like there was some recent changes in upstream here that I wasn't aware of: https://github.com/prometheus/alertmanager/pull/3225. Now that we've synced I don't think I need to do this anymore.


The empty name is expected, that's how it's defined in [FromGlob](https://github.com/grafana/prometheus-alertmanager/blob/main/template/template.go#L52-L55).

notify/templates.go
JacobsonMT: Fixed some edge cases with empty name and added tests, should have better support now.

notify/grafana_alertmanager.go
```diff
@@ -140,18 +152,19 @@ type Configuration interface {
 	DispatcherLimits() DispatcherLimits
 	InhibitRules() []InhibitRule
 	MuteTimeIntervals() []MuteTimeInterval
-	ReceiverIntegrations() (map[string][]*Integration, error)
-	BuildReceiverIntegrationsFunc() func(next *GrafanaIntegrationConfig, tmpl *templates.Template) (Notifier, error)
+	Receivers() map[string]*APIReceiver
```
yuri-tceretian: I think this can be a slice because information about receiver name is in APIReceiver.

yuri-tceretian: By doing this you will be able to remove this confusing method

https://github.com/grafana/grafana/pull/66901/files#diff-4aa6a5bc542bc278b3cddc81fe825af177f1f354c9dbb08d0f01d86b7dbe414eR313-R319

yuri-tceretian (approved):
LGTM. My last comments are nice to have but not blockers

grobinson-grafana (approved):

PR #79, opened 2023-04-26
Rename image.ImageStore to image.Provider

santihernandezc:
This PR renames our `images.ImageStore` to `images.Provider` for two main reasons:
- This interface is supposed to provide images. It does not store them, nor does it interact with the DB.
- This new naming avoids confusion with other similarly named interfaces in `grafana/grafana` like `ImageStore` (which interacts with the DB) and `ImageService`.

This interface will be refactored to support the use cases described in https://github.com/grafana/alerting-squad/issues/480

The number of modified files is pretty big, but this is just a rename, no functionality has been changed.

### DIFF
--- a/images/images.go
+++ b/images/images.go
@@ -28,13 +28,13 @@ func (i Image) HasURL() bool {
 	return i.URL != ""
 }
 
-type ImageStore interface {
+type ImageProvider interface {
 	GetImage(ctx context.Context, token string) (*Image, error)
 }
 
-type UnavailableImageStore struct{}
+type UnavailableImageProvider struct{}
 
 // GetImage returns the image with the corresponding token, or ErrImageNotFound.
-func (u *UnavailableImageStore) GetImage(context.Context, string) (*Image, error) {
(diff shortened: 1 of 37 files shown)


yuri-tceretian (approved):

images/images.go
```diff
@@ -28,13 +28,13 @@ func (i Image) HasURL() bool {
 	return i.URL != ""
 }
 
-type ImageStore interface {
+type ImageProvider interface {
```
alexweav: `images` is already in the package name, so the consumer must say `images.ImageProvider` which has some repetition.

WDYT about `images.Provider` instead?

yuri-tceretian: Oh, I like this idea! We can even split into two interfaces `images.UrlProvider` and `images.BinaryProvider` to have more specific interfaces for use-cases (many integrations just use URL and do not need method that provides binary)

santihernandezc: Oh I like that!

yuri-tceretian (approved):

alexweav (approved):
LGTM

PR #80, opened 2023-04-26
Describe methods to get an image URL/bytes

santihernandezc:
This PR adds methods to get an image URL or bytes, depending on what each integration needs. These methods are described in the `ImageStore` interface, but not yet implemented.

The next PR will use these methods to add images in notifications instead of using `utils.WithStoredImages()`. The `ImageStore.GetImage()` method will be deleted.

Related issue: https://github.com/grafana/alerting-squad/issues/480


yuri-tceretian (approved):
LGTM

gotjosh (approved):
LGTM

images/testing.go
```diff
@@ -23,6 +28,46 @@ func (f *FakeProvider) GetImage(_ context.Context, token string) (*Image, error)
...
+		}
+	}
+	return nil, ErrImageNotFound
+}
+
+// getImageURI is a helper function to retrieve the image URI from the alert annotations as a string.
+func getImageURI(alert types.Alert) (string, error) {
+	uri, ok := alert.Annotations[models.ImageTokenAnnotation]
+	if !ok {
+		return "", fmt.Errorf("no image uri in annotations")
```
gotjosh, line 66: I think you should make this a named error. I suggest `ErrNoImageURI`

Because then on the callsite, you can do things like 

```golang
	uri, err := getImageURI(alert)
        if err != nil {
	  if errors.Is(ErrNoImageURI) {
		  return nil, ErrImageNotFound
	  }
         return err
}
```

images/testing.go
```diff
@@ -23,6 +28,46 @@ func (f *FakeProvider) GetImage(_ context.Context, token string) (*Image, error)
...
+
+	for _, img := range f.Images {
+		if img.Token == uri || img.URL == uri {
+			return img.URL, nil
+		}
+	}
+	return "", ErrImageNotFound
+}
+
+// GetRawImage returns an io.Reader to read the bytes of the image associated with a given alert.
```
gotjosh, line 46: As part of the comment description, if you're returning a named error is good to let the user of this function know they can assert on the specific error. 

So my suggestion is:

```
// GetRawImage returns an io.Reader to read the bytes of the image associated with a given alert. Returns `ErrImageNotFound` if no image is found.
```

PR #81, opened 2023-04-26
Update Alertmanager to 803b73a

### LINKED ISSUE #63584: [Bug] Alerting: Malformed/garbled strings with Go functions in Slack webhook
**What happened**:
Our alert templates include some piping and Go formatting like `title` and `reReplaceAll` is formatting some strings but nothing terribly complicated. Sometimes we see totally garbled text that shouldn't appear under any circumstances. This might happen when multiple alerts are fired simultaneously?

For example, `sierra_leone` should have become `Siera Leone` but instead it either appears blank, or becomes something like `sSrrrlLOn`. Same happened with `grafana_state_error` becoming `Graf` or `GrAanaa sAAerAoo`.

**What you expected to happen**:
Expected formatted strings consistently.

**How to reproduce it (as minimally and precisely as possible)**:
I think the culprit might be a ```{{ reReplaceAll "foo" "bar" .Name | title }}``` where this replace is piped into a title? Those are the cases where I have see garbed text.

**Anything else we need to know?**:

**Environment**:
- Grafana version: 9.3.6
- Data source type & version: Varied
- OS Grafana is installed on: Linux
- User OS & Browser: N/A
- Grafana plugins: N/A
- Others: N/A

```[tasklist]

```

grobinson-grafana:
This commit updates Alertmanager to [`803b73a`](https://github.com/grafana/prometheus-alertmanager/commit/803b73a2f1f7c59bffda84fd3c3ff9ab625c9413) to fix https://github.com/grafana/grafana/issues/63584.


yuri-tceretian (approved):

grobinson-grafana:
This pull request is so we can backport the fix for https://github.com/grafana/grafana/issues/63584 to 9.5.x.

PR #82, opened 2023-04-27
Use image URL or bytes in the Discord integration

santihernandezc:
This PR uses the new methods added to `images.Provider` to get images in Discord notifications.

Discord requires a file name when uploading an image, so `GetImageBytes()` now returns the image name.

Other changes:
- Comments for methods in `images.Provider` were moved to the interface definition, deleting them from the implementation (`UnavailableProvider `).
- The `constructAttachments()` method in the Discord integration was refactored, extracting some code into separate helper methods.


images/images.go
```diff
@@ -33,8 +36,13 @@ func (i Image) HasURL() bool {
...
 	GetImage(ctx context.Context, token string) (*Image, error)
-	GetImageURL(ctx context.Context, alert types.Alert) (string, error)
-	GetRawImage(ctx context.Context, alert types.Alert) (io.Reader, error)
+
+	// GetImageURL returns the URL of an image associated with a given alert.
+	GetImageURL(ctx context.Context, alert *types.Alert) (string, error)
+
+	// GetRawImage returns an io.Reader to read the bytes of an image associated with a given alert,
+	// a string representing the filename, and an optional error.
+	GetRawImage(ctx context.Context, alert *types.Alert) (io.Reader, string, error)
```
alexweav: Is the fake/unavailable the only thing implementing this right now? I saw no other impls of `GetRawImage` in this repo or in grafana/main.

Should it return a more specific `io.ReadCloser` directly, rather than forcing everything to cast down to `io.Reader` and we lose all the `.Close()` calls? Files should still need to be closed I'd imagine.

receivers/discord/discord.go
```diff
@@ -202,52 +201,62 @@ func (d Notifier) SendResolved() bool {
...
+		// We got an attachment, either using the image URL or bytes.
+		attachments = append(attachments, attachment)
+		embedQuota--
+	}
 
 	return attachments
 }
 
+func (d Notifier) getAttachmentFromURL(ctx context.Context, alert *types.Alert) (discordAttachment, error) {
+	url, err := d.images.GetImageURL(ctx, alert)
```
alexweav, line 245: How did we locally test this change without Grafana having been updated to use the latest `alerting` version? The live image store in grafana does not have any working implementation of `GetImageURL`. Is there a separate PR or branch for that?

images/images.go
yuri-tceretian: Agree. Good idea to use io.ReadCloser.

yuri-tceretian:
We basically have 0 tests of notifiers with images. Can you add them?

santihernandezc:
@yuri-tceretian sure, that's part of the next commit, at least for Discord.

receivers/discord/discord.go
```diff
@@ -202,52 +201,62 @@ func (d Notifier) SendResolved() bool {
...
 
-	_ = images.WithStoredImages(ctx, d.log, d.images,
-		func(index int, image images.Image) error {
-			if embedQuota < 1 {
-				return images.ErrImagesDone
+		attachment, err := d.getAttachmentFromURL(ctx, alert)
+		if err != nil {
+			if !errors.Is(err, images.ErrImagesNoURL) {
+				// If there's an error, continue iterating through the other alerts.
+				continue
```
yuri-tceretian: shouldn't we log error in this case? At least with warn severity

yuri-tceretian: I would go with inverse approach

images/images.go
santihernandezc: > Is the fake/unavailable the only thing implementing this right now?

Yes, Discord would be the first integration to implement it.

> Should it return a more specific `io.ReadCloser` directly

👌

receivers/discord/discord.go
santihernandezc: > Is there a separate PR or branch for that?

Not yet, the plan was to implement that in `grafana/grafana` next.

receivers/discord/discord.go
santihernandezc: What do you mean by "inverse approach"? I agree with logging some warning if there's an error.

yuri-tceretian:
```
if errors.Is(err, images.ErrImagesNoURL) {
     // The image has no public URL, use the bytes for the attachment.
    attachment, err = d.getAttachmentFromBytes(ctx, alert)    
}
if err != nil {
   d.log.Warn(...., err)
}

File: receivers/discord/discord.go [hunk 4]
```diff
@@ -202,50 +201,65 @@ func (d Notifier) SendResolved() bool {
...
 }
 
-func (d Notifier) constructAttachments(ctx context.Context, as []*types.Alert, embedQuota int) []discordAttachment {
-	attachments := make([]discordAttachment, 0)
+func (d Notifier) constructAttachments(ctx context.Context, alerts []*types.Alert, embedQuota int) []discordAttachment {
+	attachments := make([]discordAttachment, 0, embedQuota)
+	for _, alert := range alerts {
+		// Check if the image limit has been reached at the start of each iteration.
+		if embedQuota < 1 {
+			break
```
alexweav, line 211: Could we log a warning if we hit the limit? (This line may technically not be the best spot for it, if we're looping over alerts, and some alerts don't have screenshots.) I just think we should log a warning, if we are unable to attach an image due to the limit.

receivers/discord/discord.go
```diff
@@ -202,50 +201,65 @@ func (d Notifier) SendResolved() bool {
...
+		}
 
-	_ = images.WithStoredImages(ctx, d.log, d.images,
-		func(index int, image images.Image) error {
-			if embedQuota < 1 {
-				return images.ErrImagesDone
+		attachment, err := d.getAttachmentFromURL(ctx, alert)
```
alexweav: Not a problem to solve in this PR.. but something to keep in mind, I just wanted to get your thoughts on it :smile: :

Is this a network call/database call? We're doing it in a loop, in the future I wonder if we should perform some batching here.

receivers/discord/discord.go
```diff
@@ -202,50 +201,65 @@ func (d Notifier) SendResolved() bool {
...
+		if err != nil {
+			if errors.Is(err, images.ErrNoImageForAlert) {
```
alexweav, line 216: `errors.Is` is `false` if `err` is nil, so I think this structure can be simplified a bit, and we can get rid of all else cases:
```
attachment, err := ...
if errors.Is(err, images.ErrNoImageForAlert) {
    continue
}
if errors.Is(err, ...) {
    attachment, err := d.getAttachmentFromBytes(...)
    // No continue needed! Note that we're re-using `err` from outer scope. Next check will cover it.. or if this succeeds, it gets set to nil.
    // We can leave the inner `if` if we really want here, but we don't have to.
}
if err != nil {
    d.log.Error(...)
    continue
}
```
Then, it reads more like a book, top-to-bottom, and has less nesting.

notify/grafana_alertmanager_test.go
```diff
@@ -51,33 +50,33 @@ func TestPutAlert(t *testing.T) {
 			title: "Valid alerts with different start/end set",
 			postableAlerts: amv2.PostableAlerts{
 				{ // Start and end set.
-					Annotations: models.LabelSet{"msg": "Alert1 annotation"},
```
santihernandezc, line 54: This change is the only one throughout this file, I deleted a duplicated import and replaced `models` with the alias `amv2`

images/images.go
```diff
@@ -17,7 +17,16 @@ var (
 	// the maximum number of images has been iterated.
 	ErrImagesDone = errors.New("images done")
 
+	// ErrImagesNoPath is returned whenever an image is found but has no path on disk.
+	ErrImagesNoPath = errors.New("no path for image")
```
gotjosh, line 21: This is not currently being used, shall we introduce it later on when it is actually used?

images/images.go
```diff
@@ -17,7 +17,16 @@ var (
...
+
+	// ErrImagesNoURL is returned whenever an image is found but has no URL.
+	ErrImagesNoURL = errors.New("no URL for image")
+
 	ErrImagesUnavailable = errors.New("alert screenshots are unavailable")
+
+	// ErrNoImageForAlert is returned when no image is associated to a given alert.
+	ErrNoImageForAlert = errors.New("no image for alert")
 )
```
gotjosh, line 30: nit: Some of this naming is inconsistent, we have `ErrImageNotFound` which uses singular `Image` but we also have `ErrImagesNoURL` which seems to refer to a single image. 

My suggestion is to just drop the `Image` word from all of these errors as they're already prefixed with the word `image` from the package. As such, they're used as `image.ErrNoURL`, `image.ErrNoPath` or `ErrWithNoAlert`

images/images.go
```diff
@@ -33,8 +42,13 @@ func (i Image) HasURL() bool {
 
 type Provider interface {
 	GetImage(ctx context.Context, token string) (*Image, error)
```
gotjosh, line 50: Can we also document this method?

images/images.go
```diff
@@ -33,8 +42,13 @@ func (i Image) HasURL() bool {
...
-	GetImageURL(ctx context.Context, alert types.Alert) (string, error)
-	GetRawImage(ctx context.Context, alert types.Alert) (io.Reader, error)
+
+	// GetImageURL returns the URL of an image associated with a given alert. Returns `ErrImageNotFound` if no image is found.
```
gotjosh: In the stub, this returns `ErrImagesNoURL` as opposed to `ErrImageNotFound` or perhaps this is different? Should we document the return of `ErrImagesNoURL`?

notify/grafana_alertmanager_test.go
gotjosh: Thank you for this comment!

receivers/discord/discord.go
gotjosh: nit: I don't understand the name of this method; get an attachment from an URL? But we're not giving it any URL instead we're giving it an alert...

My suggestion is to rename this to `attachmentFor(ctx, alert)`. Feels like the fact that we get an attachment with an URL is an implementation detail of the method itself and not really relevant?

receivers/discord/discord.go
```diff
@@ -202,50 +201,64 @@ func (d Notifier) SendResolved() bool {
...
+			break
+		}
 
-	_ = images.WithStoredImages(ctx, d.log, d.images,
-		func(index int, image images.Image) error {
-			if embedQuota < 1 {
-				return images.ErrImagesDone
-			}
+		attachment, err := d.getAttachmentFromURL(ctx, alert)
+		if errors.Is(err, images.ErrNoImageForAlert) {
```
gotjosh: I found this extremely hard to read - as such, here's my suggestion:

```golang
func (d Notifier) constructAttachments(ctx context.Context, alerts []*types.Alert, embedQuota int) []discordAttachment {
	attachments := make([]discordAttachment, 0, embedQuota)
	embedsUsed := 0
	for _, alert := range alerts {
		// Check if the image limit has been reached at the start of each iteration.
		if embedsUsed >= embedQuota {
			d.log.Warn("Discord embed quota reached, not creating more attachments for this notification", "embedQuota", embedQuota)
			break
		}

		attachment, err := d.getAttachment(ctx, alert)
		if err != nil {
			if errors.Is(err, images.ErrNoImageForAlert) {
				// There's no image for this alert, continue.
				continue
			}

			d.log.Error("failed to create an attachment for Discord", "alert", alert, "error", err)
			continue
		}

		// We got an attachment, either using the image URL or bytes.
		attachments = append(attachments, attachment)
		embedsUsed++
	}

	return attachments
}

// getAttachment tries to construct a discord attachment for the alert image. If the image has no public URL
// it uses the raw bytes for uploading directly to discord.
func (d Notifier) getAttachment(ctx context.Context, alert *types.Alert) (discordAttachment, error) {
	attachment, err := d.getAttachmentFromURL(ctx, alert)

	// There's an image, but it has no public URL, use the bytes for the attachment.
	if errors.Is(err, images.ErrImagesNoURL) {
		return d.getAttachmentFromBytes(ctx, alert)
	}

	return attachment, err
}

```

receivers/discord/discord.go
```diff
@@ -280,7 +293,6 @@ func (d Notifier) buildRequest(url string, body []byte, attachments []discordAtt
 	for _, a := range attachments {
 		if a.reader != nil { // We have an image to upload.
 			err = func() error {
-				defer func() { _ = a.reader.Close() }()
```
gotjosh: You still need to close the the reader if you used `getAttachmentFromBytes`.

An option here is to add a cleanup function to the attachment.

```golang
type discordAttachment struct {
	url         string
	reader      io.Reader
	name        string
	alertName   string
	state       model.AlertStatus
	cleanupFunc func()
}

func (d discordAttachment) cleanup() {
	d.cleanupFunc()
}
```

then on you `getAttachmentFromBytes` can do:

```golang
	return discordAttachment{
		url:         "attachment://" + name,
		name:        name,
		reader:      r,
		state:       alert.Status(),
		alertName:   alert.Name(),
		cleanupFunc: func() { _ = r.Close() },
	}, nil
```

and the finally you can do in 
```golang
defer func() { a.cleanup() }
```

gotjosh (commented):
LGTM in principle but please see my comments.

images/images.go
santihernandezc: It's being used in grafana/grafana

images/images.go
santihernandezc: I agree 100%, I was gonna make a follow-up PR to make these errors more homogenous, basically eliminating the "Image(s)" part and ordering these errors.

gotjosh: I'm fine with having this in a separate PR, but please file an issue now and attach it as a comment to this thread 😄

images/images.go
santihernandezc: This is going to be deleted as soon as all notifiers use the two new methods, so I thought that documenting a method that existed since this package was created right when it's gonna be deprecated didn't make much sense.
I could document it anyway if you think that's better.

receivers/discord/discord.go
gotjosh: Scratch all this - we have access to the reader directly, why not close it?

images/images.go
gotjosh: You can use golang's deprecation: https://zchee.github.io/golang-wiki/Deprecated/

```suggestion
        Deprecated: Please use either GetImageURL or GetRawImage.
	GetImage(ctx context.Context, token string) (*Image, error)
```

receivers/discord/discord.go
santihernandezc: My intention was to have two methods that we can use to create an attachment:

- One that uses the image's URL.
- One that uses the image's bytes.

I didn't mean to imply that the method should accept a URL, but rather that the function uses the URL in the returned attachment. Maybe your suggestion of having only one method is a better approach.

images/images.go
santihernandezc: I just created an issue to make these names more consistent and avoid stuttering (https://github.com/grafana/alerting/issues/88).

receivers/discord/discord.go
santihernandezc: > Feels like the fact that we get an attachment with an URL is an implementation detail of the method itself and not really relevant?

Makes sense!

receivers/discord/discord.go
santihernandezc: Thanks, I removed this when I changed the `ReadCloser` for a `Reader`, but forgot to bring this back after discarding that idea.

gotjosh (approved):
LGTM

yuri-tceretian (approved):
LGTM

PR #90, opened 2023-05-05
Update Discord integration to accept URL as secure setting

### LINKED ISSUE #40148: Provisioning: Discord Notification's URL secure setting is in fact a regular setting
**What happened**:

When provisioning Grafana, [the documentation here](https://grafana.com/docs/grafana/latest/administration/provisioning/) states that Discord's `url` parameter is a "Secure Setting", however if you provide it as a secure setting:
```yaml
notifiers:
  notifiers.yaml:
    notifiers:
      - name: "Discord Alert"
        type: discord
        uid: discord-alert
        org_id: 1
        is_default: false
        send_reminder: true
        frequency: 24h
        disable_resolve_message: false
        settings:
          avatar_url: "SOME_URL"
          content: "<@SOME_ID>"
        secure_settings:
          url: "$DISCORD_URL"
 ```
 it fails with the error `service init failed: Alert notification provisioning error: alert validation error: Could not find webhook url property in settings`
 
However, if provided as a regular setting:
```yaml
settings:
  avatar_url: "SOME_URL"
  content: "<@SOME_ID>"
  url: "$DISCORD_URL"
```
everything works.

Looking at the [Discord Alert code here](https://github.com/grafana/grafana/blob/da813877fb8b6e2796d3305927ee4017351e998b/pkg/services/alerting/notifiers/discord.go#L57), and comparing it with the Slack equivalent, it seems it is indeed just a regular setting and not a secure setting.

**What you expected to happen**:

Discord's notification setting should be a Secure Setting.

yuri-tceretian:
Provisioning documentation mentions Discord URL as secure setting, but in fact it is not.

Partially fixes https://github.com/grafana/grafana/issues/40148. This will need some changes in Grafana too

### DIFF
--- a/receivers/discord/config.go
+++ b/receivers/discord/config.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/grafana/alerting/receivers"
 	"github.com/grafana/alerting/templates"
 )
 
@@ -16,12 +17,13 @@ type Config struct {
 	UseDiscordUsername bool   `json:"use_discord_username,omitempty" yaml:"use_discord_username,omitempty"`
 }
 
-func NewConfig(jsonData json.RawMessage) (Config, error) {
+func NewConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Config, error) {
 	var settings Config
 	err := json.Unmarshal(jsonData, &settings)
 	if err != nil {
 		return Config{}, fmt.Errorf("failed to unmarshal settings: %w", err)
 	}
+	settings.WebhookURL = decryptFn("url", settings.WebhookURL)
 	if settings.WebhookURL == "" {
 		return Config{}, errors.New("could not find webhook url property in settings")
 	}
--- a/notify/receivers.go
+++ b/notify/receivers.go
@@ -402,7 +402,7 @@ func parseNotifier(ctx context.Context, result *GrafanaReceiverConfig, receiver 
 		}
 		result.DingdingConfigs = append(result.DingdingConfigs, newNotifierConfig(receiver, cfg))
 	case "discord":
-		cfg, err := discord.NewConfig(receiver.Settings)
+		cfg, err := discord.NewConfig(receiver.Settings, decryptFn)
 		if err != nil {
 			return err
 		}
--- a/receivers/discord/config_test.go
+++ b/receivers/discord/config_test.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/stretchr/testify/require"
 
+	receiversTesting "github.com/grafana/alerting/receivers/testing"
 	"github.com/grafana/alerting/templates"
 )
 
@@ -13,6 +14,7 @@ func TestNewConfig(t *testing.T) {
 	cases := []struct {
 		name              string
 		settings          string
+		secrets           map[string][]byte
 		expectedConfig    Config
 		expectedInitError string
 	}{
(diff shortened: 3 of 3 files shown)


receivers/discord/config.go
```diff
@@ -16,12 +17,13 @@ type Config struct {
...
 }
 
-func NewConfig(jsonData json.RawMessage) (Config, error) {
+func NewConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Config, error) {
 	var settings Config
 	err := json.Unmarshal(jsonData, &settings)
 	if err != nil {
 		return Config{}, fmt.Errorf("failed to unmarshal settings: %w", err)
 	}
+	settings.WebhookURL = decryptFn("url", settings.WebhookURL)
```
grobinson-grafana, line 26: Is this plain text at present? Does this mean existing Discord configurations will be broken and need configuration?

yuri-tceretian: Yes, it's plain text. The current configurations will not be broken because how `decryptFn` is implemented: if encrypted key is missing, it falls back to the plain text

grobinson-grafana (approved):

santihernandezc (approved):
LGTM 👍 
I tested the changes and it works as expected with our current format for Discord.

santihernandezc (approved):
LGTM 👍 
I tested the changes and it works as expected with our current format for Discord.

receivers/discord/config.go
alexweav: I see this doesn't break config writes for the old format, but what about reads? Will this field no longer be returned from API since it's now secure? If so, this could break how Terraform updates state. We might need to make a change in provider to read in both ways.

yuri-tceretian: The field is returned as insecure until the next write that will convert it to a secure string, and provisioning API will start responding with `[REDACTED]`. 
I checked the Terraform provider and after the update, it detects the field as requiring change every time. So, this needs to be fixed in the terraform provider.

yuri-tceretian: This PR does not break anything unless there is another change in Grafana that promotes this field as secured.

alexweav: > it detects the field as requiring change every time

This will probably break the CI on https://github.com/grafana/terraform-provider-grafana/, if it runs against `main`.

The tests have a step that validates that you get a clean apply every time. Otherwise, the tests will start failing at the next Grafana release - we should really try to fix this in the provider before this change is released.

alexweav: Specifically, without this, lots of customer CIs will start behaving oddly, as many tools surrounding terraform (like Atlantis) rely on its idempotency and ability to detect when no changes will need to be made. I think with this break, Terraform will _always_ think a change needs to be made until a new version of the provider is released.

I think we should file an issue in the provider to track it.

yuri-tceretian: This particular change does not break anything. It just let the configuration read from two sources. The writing logic is in Grafana and is not touched. API reading logic is also in Grafana.

alexweav: I'm trying to rectify:
> This particular change does not break anything. It just let the configuration read from two sources.

with:
> The field is returned as insecure until the next write that will convert it to a secure string, and provisioning API will start responding with [REDACTED].
I checked the Terraform provider and after the update, it detects the field as requiring change every time.

These statements seem contradictory to me, perhaps there is something I'm missing. If we're changing how a value is returned, and TF can't see it anymore and behaves incorrectly, that's a break. Clean applies are a requirement of Terraform - TF will no longer generate clean applies once the contact point is changed. Are we saying this PR doesn't cause this? Are you making multiple PRs as part of this change, is there another future PR that does cause this break, and we should fix Terraform to read both fields before that?

yuri-tceretian: > The field is returned as insecure until the next write that will convert it to a secure string, and provisioning API will start responding with [REDACTED].
I checked the Terraform provider and after the update, it detects the field as requiring change every time.

As stated in the PR description, this is just one part of the changes. The second part is in Grafana: 
https://github.com/grafana/grafana/compare/yuri-tceretian/discord-url-secure?expand=1 
I thought you ask how it will work in the end result because this PR does not affect read\write operation in API in any way because this code is used only during validation and applying the configuration to Alertmanager.
The second part (link to diff above) will change how read\write work and it will make TF plan to have diff.

PR #108, opened 2023-07-03
Add Sprig template functions

jquick:
The current Alerting template functions are very few and make it hard to do any kind of re-mapping/arithmetic for your templates. I found myself unable to do basic counters or uniques on lists for my alert groups.

This PR adds in the Sprig template pack which is used by many projects to add some basic useful arithmetic, dictionary, list, etc functions. You can find more info here:
http://masterminds.github.io/sprig/

This implementation adds the `alertmanager` functions after to make sure to overwrite any conflicts and keep all current templates stable.

Here is also an example of where Helm is also importing these functions:
https://github.com/helm/helm/blob/main/pkg/engine/funcs.go#L44
I have also disable the env calls due to potential security risks (same as Helm).

With this addition I was able to spin up local Grafana and create a workable template like:

```
  {{- /* look for environments */ -}}
  {{- $envs := dict -}}
  {{- range $alerts -}}
    {{- if gt (len .Labels.environment) 0 -}}
      {{- $_ := set $envs .Labels.environment 1 -}}
    {{- end -}}
  {{- end -}}
  {{- if gt (len ($envs | keys)) 0 -}}
  Multiple Environments Firing:
    {{ range ($envs | keys) }}
      {{ . }}
    {{ end }}
  {{- end -}}
```

This is really nice for our usage and would be a great addition to our team.


jquick:
@armandgrillet - Thanks for your requests on this. Is there anything I can do to help push these PRs forward?

grobinson-grafana:
I think having some Sprig functions are useful, but I'm concerned about using denylists for functions such as `env`. Are there some functions in particular that you are interested in, as perhaps we could add them ourselves?

grobinson-grafana (changes requested):
I think having some Sprig functions are useful, but I'm concerned about using denylists for functions such as `env`. Are there some functions in particular that you are interested in, as perhaps we could add them ourselves?

jquick:
The main ones I think would be helpful are:
[Integer Math Functions](http://masterminds.github.io/sprig/math.html): add, max, mul, etc.
[Lists and List Functions](http://masterminds.github.io/sprig/lists.html): list, first, uniq
[String Functions](http://masterminds.github.io/sprig/strings.html): trim, wrap, randAlpha, plural
[Date Functions](http://masterminds.github.io/sprig/date.html): now, date
[Dictionaries and Dict Functions](http://masterminds.github.io/sprig/dicts.html): get, set, dict, hasKey, pluck, dig

We could do a inclusion list from sprig instead if thats easier/safer.

grobinson-grafana:
Thanks! Would you be able to first request this feature to the [Alertmanager](https://github.com/prometheus/alertmanager) maintainers?

The reasoning for this is that Grafana uses Alertmanager, but over time Grafana has made some changes to it's version of Alertmanager that means it's not longer 100% compatible. Instead, we are working to consolidate Grafana with the upstream Alertmanager code, but that means we cannot accept changes in Grafana that are not also accepted in Alertmanager.

gotjosh:
@jquick thank you very much for your contribution - as my colleague @grobinson-grafana just said I'd love to get a sense of which of these functions you consider useful and perhaps we can consider proposing them upstream (in https://github.com/prometheus/alertmanager).

For us, it's hard to maintain the blind inclusion of these functions without a proper analysis of how common is their usage. I can appreciate the helm reference, but as you might understand these projects (helm vs alerting) have completely different purposes. 

I'm closing this for now, but feel free to let us know if we're perhaps missing something.

Issue #96, opened 2023-05-09
ruleURL should open the alert rule page and not list the rules, previously it used to open the dashboard


panikdkernel:
https://github.com/grafana/alerting/blob/793c67215ba14d6515b93707caeb7e6d9bf82189/receivers/slack/slack.go#LL290C90-L290C90

yuri-tceretian:
This seems to have been done intentionally at the beginning of unified alerting https://github.com/grafana/grafana/pull/33739. In legacy alerting it points to the dashboard panel https://github.com/grafana/grafana/blob/40feee0d17c72421bfd0e9dc5deebc83f7e3bb68/pkg/services/alerting/notifiers/slack.go#L212

panikdkernel:
Is there a way to pass it as a config var ? Maybe Something like slackTitleURL or can we modify this behaviour using notifications templates?

yuri-tceretian:
The reason for that behavior is that unified alerting notification component (Alertmanager) supports grouping alerts by labels, and therefore integration can receive multiple alerts 
https://github.com/grafana/alerting/blob/793c67215ba14d6515b93707caeb7e6d9bf82189/receivers/slack/slack.go#L286-L290

Therefore, it is not clear to which alert rule the slack message should refer to.

yuri-tceretian:
> Is there a way to pass it as a config var ? 

Unfortunately, I do not see any way you can change that URL. We can make it configurable, though and use the current behavior as the default. 

Alternative is to check if there is one alert in the group and create proper URL to the rule

panikdkernel:
> Alternative is to check if there is one alert in the group and create proper URL to the rule

Yes, currently we have single alert in the group as we migrated from the legacy alerts

I tried grouping the alerts using group overrides in notifications policy and now ruleURL value does make sense, the only way currently I see to access alert is through notification templates 

Though it would be really handy if we could also config slack title url.

jquick:
I also ran into this. We currently have all our slack alerts ungrouped so this would be amazing!

jquick:
There is a work around for this. If you format a title in the slack link format:
```
<https://myLink|myTitle>
```
It will override the `titleLink` sent with the payload on slack. You can use a template to generate this just make sure you do not have any newlines.

armandgrillet:
This is done for Slack but might happen for other messages, leaving this issue open.

PR #109, opened 2023-07-04
Allow slack GeneratorURL as link

jquick:
This PR address issue: https://github.com/grafana/alerting/issues/96

This will update the Slack responder to update the link under the following circumstances:
* A single alert is fired and it as a GeneratorURL set (this should always be the case)
* Multiple alerts are fired but they all have the same Generator URL (grouped by alertname)

### DIFF
--- a/receivers/slack/slack.go
+++ b/receivers/slack/slack.go
@@ -283,11 +283,29 @@ func handleSlackJSONResponse(resp *http.Response, logger logging.Logger) (string
 	return result.Ts, nil
 }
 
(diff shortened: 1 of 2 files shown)


grobinson-grafana (approved):
I've tested this and works well. Thank you for your contribution!

gillesdemey (approved):
LGTM – great contribution!

konrad147 (approved):
LGTM!

santihernandezc (approved):
LGTM 👍

PR #112, opened 2023-07-17
Add a new receiver for Jira Service Management

akyuznurettin:


gotjosh:
Thank you very much for your contribution @akyuznurettin.

Sadly, we need to have enough support from the community (e.g. in the form of thumbs-up on an issue) to introduce new contact points.

I suggest you start by proposing an issue in https://github.com/grafana/grafana and then if we get enough support we can bring back this implementation.

PR #127, opened 2023-10-03
Make integration configs for testing publicly available

yuri-tceretian:
This PR introduces a package public variable `AllKnownConfigsForTesting` in `notify` package. This will allow us to reuse this configuration in Grafana tests (see PR https://github.com/grafana/grafana/pull/75849)

### DIFF
--- a/notify/testing.go
+++ b/notify/testing.go
@@ -2,12 +2,34 @@ package notify
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
 	"testing"
 	"time"
 
 	"github.com/prometheus/alertmanager/types"
 
+	"github.com/grafana/alerting/receivers/alertmanager"
+	"github.com/grafana/alerting/receivers/dinding"
+	"github.com/grafana/alerting/receivers/discord"
+	"github.com/grafana/alerting/receivers/email"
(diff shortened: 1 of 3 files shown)


notify/testing.go
```diff
@@ -91,3 +113,104 @@ func (f *fakeNotifier) SendResolved() bool {
...
+		}
+		for key, value := range secrets {
+			secrets[key] = base64.StdEncoding.EncodeToString([]byte(value))
+		}
+	}
+	return &GrafanaIntegrationConfig{
+		UID:                   fmt.Sprintf("%s-uid", name),
+		Name:                  name,
+		Type:                  n.NotifierType,
+		DisableResolveMessage: true,
```
rwwiv, line 212: I think it's better for this to be deterministic, but are we losing any usefulness by always setting `DisableResolveMessage` to `true`?

rwwiv (approved):
LGTM 🚀 

Small question about the removal of the removal of the `rand` calls in `getRawNotifierConfig()`

notify/testing.go
yuri-tceretian: The default value (if it's not specified in the json) is false. However, all integrations settings in this test data are explicitly specified. So, specifying `true` is more in line with other settings.

PR #128, opened 2023-10-05
Support Upload flag in Pushover integration

yuri-tceretian:
This field was introduced since the inception of Pushover integration in the new Alerting https://github.com/grafana/grafana/pull/34371, which was a copy of the legacy alerting model. 
However, at the time of implementation, the new alerting did not support images in notifications, and they were added later.

This PR updates Pushover integration to support `uploadImages` flag.

### DIFF
--- a/receivers/pushover/pushover.go
+++ b/receivers/pushover/pushover.go
@@ -215,7 +215,7 @@ func (pn *Notifier) genPushoverBody(ctx context.Context, as ...*types.Alert) (ma
 func (pn *Notifier) writeImageParts(ctx context.Context, w *multipart.Writer, as ...*types.Alert) {
 	// Pushover supports at most one image attachment with a maximum size of pushoverMaxFileSize.
 	// If the image is larger than pushoverMaxFileSize then return an error.
-	_ = images.WithStoredImages(ctx, pn.log, pn.images, func(index int, image images.Image) error {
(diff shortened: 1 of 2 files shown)


receivers/pushover/pushover.go
```diff
@@ -175,7 +175,11 @@ func (pn *Notifier) genPushoverBody(ctx context.Context, as ...*types.Alert) (ma
 		return nil, b, fmt.Errorf("failed write the message: %w", err)
 	}
 
-	pn.writeImageParts(ctx, w, as...)
+	if pn.settings.Upload {
+		pn.writeImageParts(ctx, w, as...)
```
rwwiv, line 179: I think we should do something with the error currently being swallowed by `writeImageParts()`, at minimum logging it. WDYT?

receivers/pushover/pushover.go
```diff
@@ -242,4 +246,7 @@ func (pn *Notifier) writeImageParts(ctx context.Context, w *multipart.Writer, as
 
 		return images.ErrImagesDone
 	}, as...)
+	if err != nil {
+		pn.log.Error("failed to fetch image for the notification", "error", err)
```
rwwiv, line 250:
```suggestion
		pn.log.Error("failed to add image to notification", "error", err)
```

rwwiv (approved):
Suggested change for the log line so it's hopefully more helpful, rest LGTM 👍 Nice catch!

PR #130, opened 2023-10-11
Fix errors when testing contact points

grobinson-grafana:
This commit fixes the errors "Missing receiver" and "Missing group labels" when testing contact points:

```
ERROR[10-16|20:25:22] Missing receiver                         logger=ngalert.notifier.webhook notifierUID=
ERROR[10-16|20:25:22] Missing group labels                     logger=ngalert.notifier.webhook notifierUID=
```

These are not actual errors, and just confuse users.


gotjosh (approved):
LGTM

PR #131, opened 2023-10-25
Support for responders in OpsGenie 

yuri-tceretian:
This PR adds support for a new field "responders" to OpsGenie. Documentation https://docs.opsgenie.com/docs/alert-api#create-alert 
The config structure and handling logic are taken from vanilla Alertmanager https://github.com/prometheus/alertmanager/blob/9a8d1f976e12b325ec47b84987a78b7845738be6/notify/opsgenie/opsgenie.go#L185-L213

I improved logic around expanding of "teams" responder to make sure that user provides the correct combination of fields.

Can be tested in https://github.com/grafana/grafana/pull/77159

### DIFF
--- a/receivers/opsgenie/config.go
+++ b/receivers/opsgenie/config.go
@@ -108,6 +108,9 @@ func NewConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Confi
 				return Config{}, fmt.Errorf("responder at index [%d] has unsupported type. Supported only: %s", idx, strings.Join(SupportedResponderTypes, ","))
 			}
 		}
+		if r.Type == "teams" && r.Name == "" {
+			return Config{}, fmt.Errorf("responder at index [%d] has type 'teams' but empty name. Must be comma-separated string of names", idx)
+		}
 	}
 
 	return Config{
--- a/receivers/opsgenie/config.go
+++ b/receivers/opsgenie/config.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"text/template"
 
 	"github.com/grafana/alerting/receivers"
 	"github.com/grafana/alerting/templates"
@@ -18,6 +19,15 @@ const (
 	DefaultAlertsURL = "https://api.opsgenie.com/v2/alerts"
 )
 
+var SupportedResponderTypes = []string{"team", "teams", "user", "escalation", "schedule"}
+
+type MessageResponder struct {
+	ID       string `json:"id,omitempty" yaml:"id,omitempty"`
+	Name     string `json:"name,omitempty" yaml:"name,omitempty"`
+	Username string `json:"username,omitempty" yaml:"username,omitempty"`
+	Type     string `json:"type" yaml:"type"` // team, user, escalation, schedule etc.
+}
+
 type Config struct {
 	APIKey           string
 	APIUrl           string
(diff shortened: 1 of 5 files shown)


receivers/opsgenie/config.go
```diff
@@ -74,6 +86,33 @@ func NewConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Confi
...
+	for idx, r := range raw.Responders {
+		if r.ID == "" && r.Username == "" && r.Name == "" {
+			return Config{}, fmt.Errorf("responder at index [%d] must have at least one of id, username or name specified", idx)
+		}
+		if strings.Contains(r.Type, "{{") {
+			_, err := template.New("").Parse(r.Type)
+			if err != nil {
+				return Config{}, fmt.Errorf("responder at index [%d] type is not a valid template: %v", idx, err)
+			}
+		} else {
```
santihernandezc, line 98: This `else` causes values containing templates to bypass validation. If the value of this field is for example `{{ printf "lol" }}` the end result will be invalid but no error will be returned.

receivers/opsgenie/config.go
```diff
@@ -74,6 +86,33 @@ func NewConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Confi
 		raw.OverridePriority = &overridePriority
 	}
 
+	for idx, r := range raw.Responders {
+		if r.ID == "" && r.Username == "" && r.Name == "" {
```
santihernandezc, line 90: Reading the section for the `visibleTo` field and the examples for both `responders` and `visibleTo` in the [Opsgenie docs](https://docs.opsgenie.com/docs/alert-api#create-alert) it's not clear to me whether any of these three values are valid for each type.
- The `user` type goes with either `id` or `username`, but not with `name`
- The `team` type goes with either `id` or `name`, but not with `username`

This validation doesn't take these edge cases into account, could you confirm that this is not an issue?

receivers/opsgenie/opsgenie_test.go
```diff
@@ -250,6 +250,136 @@ func TestNotify(t *testing.T) {
...
+					{
+                        "username": "test",
+						"type": "team"
+					}
+				]
+			}`,
+			expMsgError: nil,
+		},
+		{
+			name: "Config with teams responders should be exploded",
```
santihernandezc, line 327: So PHP! 😄

santihernandezc (changes requested):
Looks pretty good! I left a few comments, the one about the templated values being the most important one.

receivers/opsgenie/config.go
yuri-tceretian: Agree with this. The main reason to support it is the upstream. So, I think that if a user goes this way and decides to use the template the validity of the result is in their hands.

santihernandezc: But why not remove the `else` and let the template be checked as any other value after it's evaluated?

receivers/opsgenie/config.go
yuri-tceretian: It looks like the `id` and `name` can be used for any types but `username` - is only for the `user` type. 

I think it makes sense to add a validation that would make sure that if username is specified then the type must be `user`. What worries me is that it would cause even bigger drift between alertmanager flavors. If you think this is ok, I will introduce it.

receivers/opsgenie/config.go
yuri-tceretian: because the result of the template expansion could be different when it is applied with actual alerts as the context. For example `{{ .CommonAnnotations.ResponderType }}`

yuri-tceretian: Please see the test https://github.com/grafana/alerting/pull/131/files#diff-0c8a173e84ce8f8f978f1403db2f1687d0cf4aa8ef8428fe5bda9607c86f5a5aR268-R271

santihernandezc: Ah, makes sense!

receivers/opsgenie/config.go
santihernandezc: If the OpsGenie API returns an error, then adding validation at the Alertmanager level is not really a breaking change, but just catching an error before it happens. I'd say double check and go for it.

santihernandezc (approved):
LGTM 👍

receivers/opsgenie/config.go
yuri-tceretian: It turns out that invalid combinations of the fields are accepted by Opsgenie but ignored. So, I do not think it makes sense to add any additional validation that does not cause the request to fail

santihernandezc: Great, thanks for double-checking :)

PR #143, opened 2023-12-15
Add MessageThreadID to Telegram configuration

th0th:
Enable setting [message_thread_id](https://core.telegram.org/bots/api#sendmessage) for Telegram notifications.

Also see https://github.com/grafana/grafana/pull/79546.

### DIFF
--- a/receivers/telegram/config.go
+++ b/receivers/telegram/config.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/grafana/alerting/receivers"
@@ -42,6 +43,18 @@ func NewConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Confi
 	}
 	if settings.Message == "" {
 		settings.Message = templates.DefaultMessageEmbed
+	}
+
+	var messageThreadId int
+	if settings.MessageThreadID != "" {
+		messageThreadId, err = strconv.Atoi(settings.MessageThreadID)
+		if err != nil {
+			return settings, errors.New("message thread id must be an integer")
+		}
+
+		if messageThreadId != int(int32(messageThreadId)) {
+			return settings, errors.New("message thread id must be an int32")
+		}
 	}
 	// if field is missing, then we fall back to the previous default: HTML
 	if settings.ParseMode == "" {
--- a/receivers/telegram/config.go
+++ b/receivers/telegram/config.go
@@ -45,14 +45,14 @@ func NewConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Confi
 		settings.Message = templates.DefaultMessageEmbed
 	}
 
-	var messageThreadId int
+	var messageThreadID int
 	if settings.MessageThreadID != "" {
-		messageThreadId, err = strconv.Atoi(settings.MessageThreadID)
+		messageThreadID, err = strconv.Atoi(settings.MessageThreadID)
 		if err != nil {
 			return settings, errors.New("message thread id must be an integer")
 		}
 
-		if messageThreadId != int(int32(messageThreadId)) {
+		if messageThreadID != int(int32(messageThreadID)) {
 			return settings, errors.New("message thread id must be an int32")
 		}
 	}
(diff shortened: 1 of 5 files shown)


yuri-tceretian:
This is not enough to make it work. You need to update the notifier to include this field to the POST request that it sends to the Telegram API [here](https://github.com/grafana/alerting/blob/c939956dbb100df78e5eb8c47d57dc36b3feec3c/receivers/telegram/telegram.go#L111-L143) 
Also, you need to update the tests.

Please update the PR description with information about the changes. Link to another PR and repository is good but not enough.

th0th:
> This is not enough to make it work. You need to update the notifier to include this field to the POST request that it sends to the Telegram API [here](https://github.com/grafana/alerting/blob/c939956dbb100df78e5eb8c47d57dc36b3feec3c/receivers/telegram/telegram.go#L111-L143) Also, you need to update the tests.
> 
> Please update the PR description with information about the changes. Link to another PR and repository is good but not enough.

I think I did the required changes, but I can't test it on my development environment. Can you have a look please?

receivers/telegram/config.go
```diff
@@ -19,6 +19,7 @@ var SupportedParseMode = map[string]string{"Markdown": "Markdown", "MarkdownV2":
 type Config struct {
 	BotToken              string `json:"bottoken,omitempty" yaml:"bottoken,omitempty"`
 	ChatID                string `json:"chatid,omitempty" yaml:"chatid,omitempty"`
+	MessageThreadID       string `json:"message_thread_id,omitempty" yaml:"message_thread_id,omitempty"`
```
yuri-tceretian, line 23: according to the [documentation](https://core.telegram.org/bots/api#sendmessage) the field is of integer type. Here we use string, which is OK. However, I would recommend adding validation in the NewConfig function that would check that the string is a valid integer. 
This is important because the wrong configuration can lead to non-deliverable notifications.

yuri-tceretian: Please add positive and negative tests for the new field.

th0th: I added tests @yuri-tceretian. Can you please have a look and let me know if they are OK? 💐

receivers/telegram/config.go
```diff
@@ -42,6 +44,18 @@ func NewConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Confi
...
+	if settings.MessageThreadID != "" {
+		messageThreadId, err = strconv.Atoi(settings.MessageThreadID)
+		if err != nil {
+			return settings, errors.New("message thread id must be an integer")
+		}
+
+		if messageThreadId != int(int32(messageThreadId)) {
+			return settings, errors.New("message thread id must be an int32")
+		}
+	}
```
yuri-tceretian:
```suggestion
	var messageThreadID int
	if settings.MessageThreadID != "" {
		messageThreadID, err = strconv.Atoi(settings.MessageThreadID)
		if err != nil {
			return settings, errors.New("message thread id must be an integer")
		}

		if messageThreadID != int(int32(messageThreadID)) {
			return settings, errors.New("message thread id must be an int32")
		}
	}
```

receivers/telegram/telegram_test.go
```diff
@@ -62,10 +64,10 @@ func TestNotify(t *testing.T) {
 		}, {
 			name: "Multiple alerts with custom template",
 			settings: Config{
-				BotToken:  "abcdefgh0123456789",
-				ChatID:    "someid",
-				Message:   "__Custom Firing__\n{{len .Alerts.Firing}} Firing\n{{ template \"__text_alert_list\" .Alerts.Firing }}",
-				ParseMode: DefaultTelegramParseMode,
```
yuri-tceretian: this does not seem to be gofmt'ed

yuri-tceretian (approved):
LGTM! Great job. I really appreciate your contribution. Please fix the above to make CI pass and then I will merge

th0th:
> LGTM! Great job. I really appreciate your contribution. Please fix the above to make CI pass and then I will merge

Thank you 💐  I fixed the formatting issue, but I think CI requires your approval to run.

alekiv:
@th0th 

Message Thread ID is optional parameter, but now UI doesn't allow to save configuration without specifying "Message Thread ID"
image

Grafana version: 10.4.0

yuri-tceretian:
I opened a bug report https://github.com/grafana/grafana/issues/84296

PR #157, opened 2024-01-31
Update Go to 1.21

grobinson-grafana:
This commit updates go.mod and drone builds to Go 1.21. It also updates golangci-lint to version 1.55.2. In this version golint is deprecated and replaced with revive.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,10 +4,10 @@ output:
 linters:
   enable:
     - goimports
-    - golint
     - gofmt
     - misspell
     - errorlint
+    - revive
 
 linters-settings:
   errcheck:
--- a/images/utils.go
+++ b/images/utils.go
@@ -21,6 +21,8 @@ type forEachImageFunc func(index int, image Image) error
 
 // getImage returns the image for the alert or an error. It returns a nil
 // image if the alert does not have an image token or the image does not exist.
+//
+//nolint:revive
 func getImage(ctx context.Context, l logging.Logger, imageProvider Provider, alert types.Alert) (*Image, error) {
 	token := getTokenFromAnnotations(alert.Annotations)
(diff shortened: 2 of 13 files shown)


.golangci.yml
```diff
@@ -4,10 +4,10 @@ output:
 linters:
   enable:
     - goimports
-    - golint
```
grobinson-grafana, line 7: `golint` is deprecated and superseded with `revive`. Lots of linter warnings with revive, which have been ignored. Issue to fix this is tracked https://github.com/grafana/alerting-squad/issues/724.

images/utils.go
```diff
@@ -21,6 +21,8 @@ type forEachImageFunc func(index int, image Image) error
 
 // getImage returns the image for the alert or an error. It returns a nil
 // image if the alert does not have an image token or the image does not exist.
+//
+//nolint:revive
```
gillesdemey, line 25: what rule is being silenced here? Can we add a nolint for a specific violation?

gillesdemey (approved):
LGTM just one comment

images/utils.go
grobinson-grafana: It doesn't like the `if else` block here:

```go
if errors.Is(err, ErrImageNotFound) || errors.Is(err, ErrImagesUnavailable) {
	return nil, nil
} else if err != nil {
	l.Warn("failed to get image with token", "token", token, "error", err)
	return nil, err
} else {
	return img, nil
}
```

and wants me to rewrite it as:

```go
if errors.Is(err, ErrImageNotFound) || errors.Is(err, ErrImagesUnavailable) {
	return nil, nil
}
if err != nil {
	l.Warn("failed to get image with token", "token", token, "error", err)
	return nil, err
}
return img, nil
```

PR #161, opened 2024-02-26
Accept template contents as strings instead of reading from disk

JacobsonMT:
### What is this feature?

Makes use of the new upstream `Template.New()` to stop passing user-defined templates to the Grafana Alertmanager by persisting them to disk.

This PR is a general improvement and simplification, but there are also some specific reasons why we might want to do this:

- Incremental step towards abstracting data layer for Grafana's Alertmanager interface away from needing disk-based storage. Next step would likely focus on `FileStore`.
- Reduces likelihood of consistency complexities on save/apply, especially in the context of future Remote AM work.
- Since `DefaultTemplateString` is now treated like other defaults and therefor consistently applied first. This ensures that user-defined templates can now rely on being able to override defaults.

### Future improvements

- Similarly, we could consider moving the existing prometheus defaults (`default.tmpl` and `email.tmpl`) to strings as well, allowing them to be more easily overridden via API/UI.
- Similar changes can be made in Mimir Alertmanager eventually as well.

### DIFF
--- a/notify/grafana_alertmanager.go
+++ b/notify/grafana_alertmanager.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
-	"sort"
 	"sync"
 	"time"
 
@@ -98,7 +97,7 @@ type GrafanaAlertmanager struct {
 	externalURL                   string
 
 	// templates contains the template name -> template contents for each user-defined template.
-	templates map[string]string
+	templates []templates.TemplateDefinition
 }
 
 // State represents any of the two 'states' of the alertmanager. Notification log or Silences.
@@ -144,7 +143,7 @@ type Configuration interface {
 	BuildReceiverIntegrationsFunc() func(next *APIReceiver, tmpl *templates.Template) ([]*Integration, error)
 
 	RoutingTree() *Route
-	Templates() map[string]string
+	Templates() []templates.TemplateDefinition
 
 	Hash() [16]byte
 	Raw() []byte
@@ -185,7 +184,6 @@ func NewGrafanaAlertmanager(tenantKey string, tenantID int64, config *GrafanaAle
 		Metrics:           m,
 		tenantID:          tenantID,
 		externalURL:       config.ExternalURL,
-		templates:         make(map[string]string),
 	}
 
 	if err := config.Validate(); err != nil {
@@ -348,15 +346,16 @@ func (am *GrafanaAlertmanager) buildTimeIntervals(timeIntervals []config.TimeInt
 func (am *GrafanaAlertmanager) ApplyConfig(cfg Configuration) (err error) {
 	am.templates = cfg.Templates()
 
+	seen := make(map[string]struct{})
 	tmpls := make([]string, 0, len(am.templates))
 	for _, tc := range am.templates {
-		tmpls = append(tmpls, tc)
+		if _, ok := seen[tc.Name]; ok {
+			level.Warn(am.logger).Log("msg", "template with same name is defined multiple times, skipping...", "template_name", tc.Name)
+			continue
+		}
+		tmpls = append(tmpls, tc.Template)
+		seen[tc.Name] = struct{}{}
 	}
-
-	// Simple sort for consistency in template rendering. Could improve later by sorting based on the template name or defined order.
-	sort.Slice(tmpls, func(i, j int) bool {
-		return tmpls[i] < tmpls[j]
-	})
 
 	tmpl, err := am.TemplateFromContent(tmpls)
 	if err != nil {
--- a/notify/grafana_alertmanager.go
+++ b/notify/grafana_alertmanager.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"sort"
 	"sync"
 	"time"
 
@@ -351,6 +352,11 @@ func (am *GrafanaAlertmanager) ApplyConfig(cfg Configuration) (err error) {
 	for _, tc := range am.templates {
 		tmpls = append(tmpls, tc)
 	}
+
+	// Simple sort for consistency in template rendering. Could improve later by sorting based on the template name or defined order.
+	sort.Slice(tmpls, func(i, j int) bool {
+		return tmpls[i] < tmpls[j]
+	})
 
 	tmpl, err := am.TemplateFromContent(tmpls)
 	if err != nil {
--- a/notify/grafana_alertmanager.go
+++ b/notify/grafana_alertmanager.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
-	"path/filepath"
 	"sync"
 	"time"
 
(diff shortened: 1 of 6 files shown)


templates/template_data.go
```diff
@@ -26,6 +26,13 @@ type Data = template.Data
 
 var FromGlobs = template.FromGlobs
 
+type TemplateDefinition struct {
+	// Name of the template. Used to identify the template in the UI and when testing.
+	Name string
+	// Template string that contains the template definition.
```
grobinson-grafana: There is kind of a recursive definition going on here between the name of the struct and the comment. 

```suggestion
	// Template string that contains the template text.
```

templates/template_data.go
```diff
@@ -57,6 +64,30 @@ type ExtendedData struct {
 	ExternalURL string `json:"externalURL"`
 }
 
+// FromContent calls Parse on all provided template content and returns the resulting Template. Content equivalent to templates.FromGlobs.
+func FromContent(templateContents []string, options ...template.Option) (*Template, error) {
```
grobinson-grafana: I think we can use `tmpls` as it's become idiomatic to mean template text (or slice of template texts):

```suggestion
func FromContent(tmpls []string, options ...template.Option) (*Template, error) {
```

templates/template_data.go
```diff
@@ -57,6 +64,30 @@ type ExtendedData struct {
...
+	// Create new template with only file-based defaults. Done this way to simplify transition to FromContent.
+	t, err := FromGlobs(nil, options...)
```
grobinson-grafana: I think it's fine to remove `FromGlobs` in this PR, then we don't need to have a no-op `FromGlobs`.

notify/templates_test.go
```diff
@@ -102,7 +102,7 @@ func TestTemplateSimple(t *testing.T) {
 				Kind: ExecutionError,
 				Error: template.ExecError{
 					Name: "slack.title",
-					Err:  errors.New(`template: slack.title:1:38: executing "slack.title" at <{{template "missing" .}}>: template "missing" not defined`),
```
grobinson-grafana, line 105: The `slack.title` in the error message came from `FromGlobs`?

grobinson-grafana (approved):
LGTM! A couple of comments.

notify/templates_test.go
JacobsonMT: No, the extra `slack.title` in the error came from the `TestTemplate` code, it was parsing slightly differently than "real" code it was trying to mimic. Note the removed `_, err = newTextTmpl.New(c.Name).Parse(c.Template)`.

Now that we can parse templates directly, we use the same methods so it should always be inline with the actual error that would happen when ran via notifications.

templates/template_data.go
JacobsonMT: It's not a no-op as it parses the default file-based templates `default.tmpl` and `email.tmpl`, but I can remove it in the PR.

JacobsonMT: 575da9a00f98e0eb2ecafab1a82c8b5879bf9c0d

templates/template_data.go
```diff
@@ -57,6 +65,44 @@ type ExtendedData struct {
...
+// FromContent calls Parse on all provided template content and returns the resulting Template. Content equivalent to templates.FromGlobs.
+func FromContent(tmpls []string, options ...template.Option) (*Template, error) {
+	t, err := New(options...)
+	if err != nil {
+		return nil, err
+	}
+
+	// Parse default template strings defined as files. Copied from template.FromGlobs.
+	// TODO: Consider removing this and defining them as strings. This would be a breaking change for those who defined custom default files.
+	defaultFileBasedTemplates := []string{"default.tmpl", "email.tmpl"}
```
yuri-tceretian: I most likely miss something but the code below reads from hardcoded (in-memory) assets, not hard disk. `/templates/default.tmpl` is actually https://github.com/grafana/prometheus-alertmanager/blob/4f0bdf4940594319336a1faa1126c4f4e11767f8/asset/assets_vfsdata.go#L163-L168

yuri-tceretian: Therefore, if I read the code correctly, the "for" clause below can be deleted if we do not use the default templates at all, which I think is true in Grafana Managed Alerts but not in Mimir, and therefore, it can be considered as breaking change for unification.

JacobsonMT: Ah, no you didn't miss anything, got mixed up.  I'll remove the TODO entirely as I was confusing it with something else.

yuri-tceretian (approved):
LGTM.

I do not think that there is anything else to do because we do not let users to override the files on the disc. According to the code https://github.com/grafana/grafana/blob/a564c8c4398fe2a763712fa54e21885017151b45/pkg/services/ngalert/notifier/config.go#L25-L54 we always override the content of the files.

notify/grafana_alertmanager.go
```diff
@@ -96,10 +95,9 @@ type GrafanaAlertmanager struct {
 	// buildReceiverIntegrationsFunc builds the integrations for a receiver based on its APIReceiver configuration and the current parsed template.
 	buildReceiverIntegrationsFunc func(next *APIReceiver, tmpl *templates.Template) ([]*Integration, error)
 	externalURL                   string
-	workingDirectory              string
 
-	// templates contains the filenames (not full paths) of the persisted templates that were used to construct the current parsed template.
-	templates []string
+	// templates contains the template name -> template contents for each user-defined template.
+	templates []templates.TemplateDefinition
```
gotjosh, line 100: nit: Is this better off as a map? I don't see a reason for introducing another type.

notify/templates.go
```diff
@@ -74,30 +73,31 @@ func (am *GrafanaAlertmanager) TestTemplate(ctx context.Context, c TestTemplates
...
 
-	// Recreate the current template without the definition blocks that are being tested. This is so that any blocks that were removed don't get defined.
-	paths := make([]string, 0)
-	for _, name := range am.templates {
-		if name == c.Name {
-			// Skip the existing template of the same name as we're going to parse the one for testing instead.
+	// Recreate the current template replacing the definition blocks that are being tested. This is so that any blocks that were removed don't get defined.
+	var found bool
+	templateContents := make([]string, 0, len(am.templates)+1)
+	for _, td := range am.templates {
```
gotjosh, line 79: nit: I think my map suggestion would simplify this code. right?

templates/template_data.go
```diff
@@ -24,7 +25,14 @@ type Template = template.Template
 type KV = template.KV
 type Data = template.Data
 
-var FromGlobs = template.FromGlobs
+var New = template.New
```
gotjosh: I haven't looked - but want to make sure I ask the question, do we need this export?

gotjosh (approved):
LGTM - my comments are nits and I don't need to look at this again.

notify/grafana_alertmanager.go
JacobsonMT: I originally implemented as a map, but looking at GrafanaAlertmanager as a library I felt that being able to choose and guarantee ordering when passing in templates was best. It also happens to mimic the structure of the existing [TemplateDesc](https://github.com/grafana/mimir/blob/bc34d2d8ee4b7b2dfc57cb0a828188fba59ebe74/pkg/alertmanager/alertspb/alerts.pb.go#L88) proto in Mimir which, while not necessary, is nice.

notify/templates.go
JacobsonMT: See above [comment](https://github.com/grafana/alerting/pull/161#discussion_r1509355613)

PR #165, opened 2024-03-06
Remove object matchers from GettableStatus

grobinson-grafana:
This commit removes object matchers from `GettableStatus` as the Grafana UI now supports UTF-8 in matchers.


yuri-tceretian (approved):
LGTM

PR #169, opened 2024-03-19
Expose prom-AM silence state struct and helper methods

JacobsonMT:
### What is this feature?

Exposes upstream prometheus alertmanager silence [state](https://github.com/prometheus/alertmanager/blob/3ee2cd0f1271e277295c02b6160507b4d193dde2/silence/silence.go#L908) type and some related helper methods, such as [DecodeState](https://github.com/prometheus/alertmanager/blob/3ee2cd0f1271e277295c02b6160507b4d193dde2/silence/silence.go#L942) and [MarshalBinary](https://github.com/prometheus/alertmanager/blob/3ee2cd0f1271e277295c02b6160507b4d193dde2/silence/silence.go#L931).

These methods will be used by `grafana/grafana` [PR](https://github.com/grafana/grafana/pull/84705) to persist the silence state to the kvstore immediately instead of waiting for the
next maintenance run. This is used after Create/Delete to prevent silences from being lost when a new Alertmanager is started before the state has persisted. This can happen, for example, in a rolling deployment scenario when HA is not enabled.

### Considerations of note

- A trade-off between type safety and performance is made in `SilenceState() (SilenceState, error)`. Owing to the fact that upstream prom-AM does not expose the necessary types / methods, we marshalled the silence just to unmarshalling it again. Alternateively, we could keep the return value as (string, error) and return the string directly as, for now, callers just needs the string itself. However, this would remove type safety on the interface, forcing the caller to trust the AM implementation to always return a consistent type or to perform the unmarshalling themselves. Suggestions welcome, but keep in mind the `grafana/grafana` implementation and how it might interact with remote alertmanager.

### DIFF
--- a/notify/silences.go
+++ b/notify/silences.go
@@ -1,11 +1,15 @@
 package notify
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"time"
 
 	"github.com/go-kit/log/level"
+	"github.com/matttproud/golang_protobuf_extensions/pbutil"
+
 	v2 "github.com/prometheus/alertmanager/api/v2"
 	amv2 "github.com/prometheus/alertmanager/api/v2/models"
 	"github.com/prometheus/alertmanager/silence"
@@ -151,4 +155,55 @@ func (am *GrafanaAlertmanager) DeleteSilence(silenceID string) error {
 	}
 
 	return nil
+}
+
+func (am *GrafanaAlertmanager) SilenceState() (SilenceState, error) {
+	r, w := io.Pipe()
+	go func() {
+		_, err := am.silences.Snapshot(w)
+		_ = w.CloseWithError(err)
+	}()
+
+	// Trade-off between type safety and performance ahead.
+	// This is a bit awkward as we marshalled the silence just to unmarshalling it again. We could keep the return value
+	// as (string, error) and return the string directly as, for now, callers just needs the string itself.
+	/// However, this would remove type safety on the interface, forcing the caller to trust the AM implementation to
+	// always return a consistent type or to perform the unmarshalling themselves.
+	return DecodeState(r)
+}
+
+// SilenceState copied from state in prometheus-alertmanager/silence/silence.go.
+type SilenceState map[string]*silencepb.MeshSilence
+
+func (s SilenceState) MarshalBinary() ([]byte, error) {
+	var buf bytes.Buffer
+
+	for _, e := range s {
+		if _, err := pbutil.WriteDelimited(&buf, e); err != nil {
+			return nil, err
+		}
+	}
+	return buf.Bytes(), nil
+}
+
+// DecodeState copied from decodeState in prometheus-alertmanager/silence/silence.go.
+func DecodeState(r io.Reader) (SilenceState, error) {
+	st := SilenceState{}
+	for {
+		var s silencepb.MeshSilence
+		_, err := pbutil.ReadDelimited(r, &s)
+		if err == nil {
+			if s.Silence == nil {
+				return nil, silence.ErrInvalidState
+			}
+			st[s.Silence.Id] = &s
+			continue
+		}
+		//nolint:errorlint
+		if err == io.EOF {
+			break
+		}
+		return nil, err
+	}
+	return st, nil
 }
--- a/notify/silences.go
+++ b/notify/silences.go
@@ -199,8 +199,7 @@ func DecodeState(r io.Reader) (SilenceState, error) {
 			st[s.Silence.Id] = &s
 			continue
 		}
-		//nolint:errorlint
-		if err == io.EOF {
+		if errors.Is(err, io.EOF) {
 			break
 		}
 		return nil, err
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,8 @@ require (
 	github.com/benbjohnson/clock v1.3.5
 	github.com/go-kit/log v0.2.1
 	github.com/go-openapi/strfmt v0.22.0
+	github.com/google/go-cmp v0.6.0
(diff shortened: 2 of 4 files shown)


notify/silences.go
```diff
@@ -124,3 +129,54 @@ func (am *GrafanaAlertmanager) DeleteSilence(silenceID string) error {
...
+		_, err := pbutil.ReadDelimited(r, &s)
+		if err == nil {
+			if s.Silence == nil {
+				return nil, silence.ErrInvalidState
+			}
+			st[s.Silence.Id] = &s
+			continue
+		}
+		//nolint:errorlint
+		if err == io.EOF {
```
grobinson-grafana: This should be `errors.Is`? https://github.com/prometheus/alertmanager/blob/main/silence/silence.go#L964

notify/silences.go
```diff
@@ -124,3 +129,54 @@ func (am *GrafanaAlertmanager) DeleteSilence(silenceID string) error {
 
 	return nil
 }
+
+func (am *GrafanaAlertmanager) SilenceState() (SilenceState, error) {
+	r, w := io.Pipe()
```
grobinson-grafana, line 161: There are a couple of things here that I don't understand:

1. How come we snapshot the silence in a goroutine?
2. What's the reason behind using `io.Pipe` instead of `bytes.Buffer`?

notify/silences.go
JacobsonMT: Huh, nice catch! Guess we updated the AM sometime between when I copied that code and now.

notify/silences.go
JacobsonMT: This part is not necessary, I originally had a `bytes.Buffer` but noticed this nifty `io.Pipe` making it so it doesn't have to fully buffer the snapshot in memory. This is the standard way of using it IIUC, but I can put back the `bytes.Buffer` if that's preferred as I see it's not heavily used in Grafana in general.

grobinson-grafana: If `io.Pipe` is more performant then I'm 👍

grobinson-grafana (approved):

gotjosh:
Nice work!

> This is used after Create/Delete to prevent silences from being lost when a new Alertmanager is started before the state has persisted

I'm not sure about this approach - wouldn't this create a write amplification problem that we're trying to get rid of?

I think eventual consistency is OK for this case as long we think the 15m (IIRC maintenance is every 15m) consistency window is OK.

In my head, this is the scenario in the current codebase:

- replica 0 snapshots at t1
- replica 1 starts at t2
- replica 0 receives a new silence
- replica 0 shutdowns and takes a snapshot (there's now a divergence between what's stored and what replica 1 has running)
- replica 1 saves the current running state (bug occurs - we've deleted the silence created in replica 0 because replica 1 does not check for the current state)

I think that if we change that last step to make sure you check the database before the maintenance (and merge both states if there's divergence - there already helpers I think to merge two states) - I think this would actually be an improvement to what exists today where we execute a blind write and we could only write if there's a change on any end.

Pair that up with a new setting (not sure if there's one already) to expose the configuration of maintenance interval (for those customers that would like faster consistency)  and you should cover most cases.

Be mindful that what I'm writing above is not perfect, and it's still subject to subtle bugs (what happens if we create the same silence before and after the restart and then merge states? Would eventual consistency surprise the user?) but it's a pragmatic approach to a problem that only:

- Manifest itself is you're not using HA.
- You need to time it very perfectly e.g. creating a silence, no maintenance and a restart
- In the Cloud Alertmanager, we run 3 replicas by default. This problem won't exist after the migration.

grobinson-grafana:
I think we can offer an even better alternative for Grafana OSS. This  is not relevant for the Cloud AM, but that is a separate concern.

Instead of using the maintenance function to make the database consistent with the silences that are in-memory, we can do the opposite. We use the maintenance function to make what's in-memory consistent with what's in the database. This also means Grafana OSS can have strong consistency for silences.

This is a positive change as silences are the ONLY eventually consistent entity we have. No other entity – alert rules, contact points, notification policies and mute timings – can be lost after the UI says they are created – even if Grafana crashes. Not silences though. I think that's really bad UX.

Removing old silences is simple too. You just run a query that deletes expired silences from the table older than X.

gotjosh:
Consider my comment non-blocking - if you two agree wit this approach, I'd say let's ship it.

Just wanted to make sure the change we're making here is conscious of its implications as I don't see them mentioned.

notify/silences.go
santihernandezc: `io.Pipe()` uses channels for both the reader and the writer, so calling `am.silences.Snapshot(w)` in the same goroutine would block indefinitely.

santihernandezc (approved):
I'm fine with either approach, checking + merging before saving or strong consistency by persisting state right away.
If we see issues with the amount of writes we can always change our approach, but in principle it LGTM 👍

JacobsonMT:
> Instead of using the maintenance function to make the database consistent with the silences that are in-memory, we can do the opposite. We use the maintenance function to make what's in-memory consistent with what's in the database. This also means Grafana OSS can have strong consistency for silences.
>
> This is a positive change as silences are the ONLY eventually consistent entity we have. No other entity – alert rules, contact points, notification policies and mute timings – can be lost after the UI says they are created – even if Grafana crashes. Not silences though. I think that's really bad UX.
> 
> Removing old silences is simple too. You just run a query that deletes expired silences from the table older than X.

This was effectively my thinking as well, there are a bit more details in the [sister PR](https://github.com/grafana/grafana/pull/84705) where I mention eventually moving toward CRD model. For now, this is mostly meant to move silence CRUD closer to how our other resources behave. Once we align on a db structure for silences (which will likely be discussed during v2 silence API design) we can do this next iteration.

PR #170, opened 2024-03-19
Load initial silence and nflog state from memory instead of disk

JacobsonMT:
### What is this feature?

Currently, the initial loading of silence state and notification log state reads from disk. This change replaces the disk read via `SnapshotFile` with a string reader via `SnapshotReader`.

Followup to https://github.com/grafana/alerting/pull/161 to remove another unnecessary disk-usage. Silence state and nflog are already persisted and read from the kvstore. The file on disk is used only on initial load and is otherwise redundant.

### Considerations of note

- Fairly simple change as `SnapshotReader` already exists on upstream prom-AM for this exact use-case.
- We still need to pass in a snapshot filename placeholder to the maintenance function even though it won't be used. Otherwise, the [state won't be persisted on shutdown](https://github.com/prometheus/alertmanager/blob/3ee2cd0f1271e277295c02b6160507b4d193dde2/silence/silence.go#L436).

### DIFF
--- a/notify/grafana_alertmanager.go
+++ b/notify/grafana_alertmanager.go
@@ -5,11 +5,13 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+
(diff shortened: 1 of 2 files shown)


yuri-tceretian (approved):
LGTM

PR #171, opened 2024-03-19
Create silences with pre-defined IDs

santihernandezc:
This PR updates our Alertmanager fork to include https://github.com/grafana/prometheus-alertmanager/pull/67 and uses the `silences.Upsert()` method. We need to be able to create silences with pre-defined IDs to replicate silences between Alertmanagers (internal, remote).

Note: we need to merge https://github.com/grafana/prometheus-alertmanager/pull/67 before this PR. Once that's done, I can update go.mod and continue.


notify/silences.go
```diff
@@ -101,12 +101,9 @@ func (am *GrafanaAlertmanager) CreateSilence(ps *PostableSilence) (string, error
 		return "", fmt.Errorf("%s: %w", msg, ErrCreateSilenceBadPayload)
 	}
 
-	silenceID, err := am.silences.Set(sil)
+	silenceID, err := am.silences.Upsert(sil)
```
gotjosh: I think we should create a new `UpsertSilence` -- This way we can have the individual mode in Grafana use whatever mechanism it requires.

notify/silences.go
```diff
@@ -82,35 +83,58 @@ func (am *GrafanaAlertmanager) GetSilence(silenceID string) (GettableSilence, er
...
+
+	return silenceID, nil
+}
+
+// marshalSilence converts a PostableSilence to a Silence protobuf and validates it.
+func (am *GrafanaAlertmanager) marshalSilence(ps *PostableSilence) (*silencepb.Silence, error) {
 	sil, err := v2.PostableSilenceToProto(ps)
 	if err != nil {
 		level.Error(am.logger).Log("msg", "marshaling to protobuf failed", "err", err)
-		return "", fmt.Errorf("%s: failed to convert API silence to internal silence: %w",
```
gotjosh: nit: you could inline `PostableSilenceToProto` and then just call this function validate silence - feels more natural than calling a function `marshalSilence` to convert to protobuf (typically these are suffixed with `ToProto`) *AND* valiting it.

gotjosh (approved):
LGTM - I don't need to see this again.

PR #179, opened 2024-04-09
Make PostableApiAlertingConfig.Validate() public

santihernandezc:
This PR makes the `validate()` method public to be able to use it even if `UnmarshalJSON()` is not being called.

Related Mimir PR: https://github.com/grafana/mimir/pull/7858


yuri-tceretian (approved):
LGTM.

PR #182, opened 2024-04-23
Drop rule URL from LINE notifier message body 

nellaG:
The rule URL is stored and delivered under a different key through a response when using many other notifiers; however, when LINE is used as the notifier, the rule URL is included directly in the message body.
Since I mainly use LINE as the alert notifier in my project, it is impossible to customize the message body completely using a message template.
This PR drops the rule URL from LINE message body so the message looks like:

from this
image

to this
image

### DIFF
--- a/receivers/line/line.go
+++ b/receivers/line/line.go
@@ -42,7 +42,7 @@ func New(cfg Config, meta receivers.Metadata, template *templates.Template, send
 func (ln *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 	ln.log.Debug("executing line notification", "notification", ln.Name)
 
-	body := ln.buildMessage(ctx, as...)
+	body := ln.buildLineMessage(ctx, as...)
 
 	form := url.Values{}
 	form.Add("message", body)
@@ -69,7 +69,7 @@ func (ln *Notifier) SendResolved() bool {
 	return !ln.GetDisableResolveMessage()
 }
 
-func (ln *Notifier) buildMessage(ctx context.Context, as ...*types.Alert) string {
+func (ln *Notifier) buildLineMessage(ctx context.Context, as ...*types.Alert) string {
 	ruleURL := path.Join(ln.tmpl.ExternalURL.String(), "/alerting/list")
 
 	var tmplErr error
--- a/receivers/line/line.go
+++ b/receivers/line/line.go
@@ -14,6 +14,7 @@ import (
 
 var (
 	// APIURL of where the notification payload is sent. It is public to be overridable in integration tests.
+	// API document link: https://notify-bot.line.me/doc/en/
 	APIURL = "https://notify-api.line.me/api/notify"
 )
 
--- a/receivers/line/line.go
+++ b/receivers/line/line.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net/url"
-	"path"
 
 	"github.com/prometheus/alertmanager/types"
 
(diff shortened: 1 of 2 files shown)


nellaG:
Requesting for review to @yuri-tceretian hope that I mentioned right reviewer

yuri-tceretian:
Hi @nellaG, thanks for contributing to the project! I appreciate it. 
I am not familiar with LINE APIs for sending messages. However, the main idea of including the URL in the message is that the notification could contain information that leads back to the rule that generated the notification. Some integrations also have links to quickly act on the alert, e.g. silence it. 
In your example, the URL does look cumbersome in the message. However, I wonder if we can find some alternative to simple deletion of the URL. If you're familiar with LINE messaging, I wonder if there are ways to add some actionable items to the message like buttons or at least support for rich text?

nellaG:
Hi @yuri-tceretian, thank you for reviewing my PR.
I already checked LINE API before making this PR, but unfortunately [they saying that](https://notify-bot.line.me/doc/en/) the message only supports text or image message. Neither button nor rich text is supported

yuri-tceretian:
@nellaG, thank you for checking! I have brought this problem to the team and we will discuss it soon. 

In the meantime: 

- Can you please include the link to the LINE API to the top of the file? This may help us in the future. 

- Also, we need to check for message length, which is 1000 characters. I would greatly appreciate if you could do that!

- The documentation does not mention whether it supports rich text or some kind of markdown. Can you check how HTML tags are rendered? Perhaps try markdown and bbcode? This will help decide in favor of removing the URL at all

nellaG:
@yuri-tceretian Ok. I'll do these and add more commits. Thank you for your support.

yuri-tceretian:
We've discussed this and decided to approve the change because:
- the URL leads to the list of rules, which does not carry any useful information
- the default message template includes GeneratorURL that points to individual rule UID, which is more useful.
- the URL cannot be deleted by template.

So, please go ahead with the changes I requested above and I will approve it.

yuri-tceretian:
Also, it seems there is great potential for improving this integration. Currently, from what I understand, it uses a notification bot that has limited functionality. However, the LINE platform's [Messaging API ](https://developers.line.biz/en/reference/messaging-api/) offers greater functionality. If the integration is changed to support custom bots, it will provide access to that API and will allow users to create nice messages with action buttons, URLs, and support for panel screenshots. 

@nellaG would you be willing to do this?

nellaG:
Hi @yuri-tceretian , thanks for your fast review of discussion about this PR. I'll add some commits (e.g. add API link into the file) and make another PR soon to improve this LINE integration alerting using message API which you suggested. I'm glad to know that there's  better messaging API of LINE because of you.

nellaG:
@yuri-tceretian I added some commits for your requirements. regarding https://github.com/grafana/alerting/pull/182/commits/5af5f2319f8b0c68c8c37d5d3304ee6c06b4fc34 , I'm unsure if I correctly captured your words.

receivers/line/line.go
```diff
@@ -42,7 +42,7 @@ func New(cfg Config, meta receivers.Metadata, template *templates.Template, send
 func (ln *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 	ln.log.Debug("executing line notification", "notification", ln.Name)
 
-	body := ln.buildMessage(ctx, as...)
+	body := ln.buildLineMessage(ctx, as...)
```
yuri-tceretian: can you add something like this below? 
https://github.com/grafana/alerting/blob/ec2c02267fa591e7252ddb44143555bc941ce6fa/receivers/telegram/telegram.go#L119-L127

nellaG: ahh the comment was about the limitation of message length! I applied this and rebased last commit 😃

yuri-tceretian: sorry for not being crystal clear!

nellaG:
@yuri-tceretian message limitation review is resolved. then is everything alright? 😃

yuri-tceretian (approved):
LGTM!

yuri-tceretian:
Sorry, the PR was lost among other things! Thank you for contributing!

PR #189, opened 2024-05-16
Add missing yaml struct tags for configuration structs

santihernandezc:
Some fields are not being unmarshalled correctly when data is in yaml format, the default behavior for yaml and json unmarshallers are different (yaml takes the lowercased field name for a struct, see [here](https://github.com/go-yaml/yaml/blob/7649d4548cb53a614db133b2a8ac1f31859dda8c/yaml.go#L392-L396)). This PR adds `yaml` tags for configuration-related structs.


JohnnyQQQQ (approved):
LGTM

PR #191, opened 2024-05-27
Rework receivers status gathering to not need upstream changes.

stevesg:
The Grafana version of the Alertmanager /api/v1/receivers API is extended
to return information about the last notification attempt for each integration.
This has been implemented as a change to the Prometheus Alertmanager fork. We
would like to avoid divergence from upstream where possible.

This change reworks the status gathering for receivers such that we do not need
upstream changes. It instead works by wrapping the notify.Notifier interface
for each integration, and capturing the necessary details. To make this as
clean as possible, there are two drop-in replacements for Receiver and
Integration, which wrap the upstream types plus the additional functionality.
This means the code that uses these types can stay largely untouched.

### DIFF
--- a/notify/grafana_alertmanager.go
+++ b/notify/grafana_alertmanager.go
@@ -13,6 +13,7 @@ import (
 	"github.com/go-kit/log/level"
 
 	"github.com/grafana/alerting/cluster"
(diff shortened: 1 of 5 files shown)


santihernandezc (approved):
LGTM 👍

PR #193, opened 2024-05-29
Validate upstream receiver configs and add missing receivers to existing checks

santihernandezc:
This PR adds validation for upstream receivers. It also adds some missing receivers to the ones we check for in case we have Grafana receivers (e.g. having Grafana receivers + an upstream Discord receiver is currently valid).

Related PR: https://github.com/grafana/mimir/pull/8200


santihernandezc:
Already merged as part of https://github.com/grafana/alerting/pull/194

PR #198, opened 2024-06-05
fix(telegram): parse correctly ChatID and MessageThreadID

### LINKED ISSUE #69950: Product-Area-Here: short description of bug here
### 
# What went wrong?

**What happened**:

In configuration file `etc/grafana/provisioning/alerting/default.yml`

the Point of Contact for the telegram is declared, the ChatId value is taken from the environment variables.

```
contactPoints:
   - orgId: 1

     name: Telegram
     receivers:
       - uid: telegram_notifier
         type: telegram
         settings:
           bottoken: $TELEGRAM_TOKEN
           chatid: $TELEGRAM_CHAT_ID

```

env variable value
`TELEGRAM_CHAT_ID="-808008888"`

When starting grafana, an error pops up:
`Error: ✗ failure to map file default.yml: failure parsing contact points: Telegram: failed to validate receiver of type "telegram": failed to unmarshal settings: json: cannot unmarshal number into Go struct field Config.chatid of type string`

The problem is not solved, even if you wrap the value in quotes:
chatid: "$TELEGRAM_CHAT_ID"

The problem was solved temporarily through a crutch:
env variable value `TELEGRAM_CHAT_ID="808008888"`
```
in config
chatid: |-
   -
   $TELEGRAM_CHAT_ID
```

The problem is that when substituting a value from an environment variable, it is interpreted as a negative number, and not as a string, which the structure expects.

### 
# How do we reproduce it?

In configuration file `etc/grafana/provisioning/alerting/default.yml`

```
contactPoints:
   - orgId: 1

     name: Telegram
     receivers:
       - uid: telegram_notifier
         type: telegram
         settings:
           bottoken: $TELEGRAM_TOKEN
           chatid: $TELEGRAM_CHAT_ID
```

where `$TELEGRAM_CHAT_ID` is number.

### 
# What Grafana version are you using?

v9.5.2

### 
### Is the bug inside a Dashboard Panel?

Copy the panel's ["get-help" data](https://grafana.com/docs/grafana/latest/troubleshooting/send-panel-to-grafana-support/) here

### Grafana Platform?

Docker

### User's OS?

_No response_

### User's Browser?

_No response_

### Is this a Regression?

None

### Are Datasources involved?

_No response_

### Anything else to add?

_No response_

NefixEstrada:
With this commit, the ChatID and MessageThreadID configuration of the Telegram notifier now can be configured either with a string or an integer, without any breaking changes for anyone consuming this library or the end users using it in their configuration files.

Fixes grafana/grafana#69950

# Reasoning

The [Telegram API specification](https://core.telegram.org/bots/api/#sendmessage) tells us that ChatID can be an Integer or a String and MessageThreadID must be an integer.

<hr>

In the PR grafana/grafana#63085, Grafana started to interpret correctly numbers and booleans from environment variables when provisioning files. So before the change, the following got evaluated:

`ENV="1"` -> `"1"` (string)

And after the change it evaluated as:

`ENV="1"` -> `1` (integer)

<hr>

With this new change, it is currently impossible to provision Telegram's ChatID or MessageThredID, since they're integers, and get parsed as such, like so:

`TELEGRAM_CHAT_ID="-1234"` -> `-1234`

And then grafana (using this library to parse the Telegram configuration), throws the following error:

```
failure parsing contact points: my-telegram: failed to validate integration "my-telegram" (UID telegram-1) of type "telegram": failed to unmarshal settings: json: cannot unmarshal number into Go struct field Config.chatid of type string
```

# Implementation details

In order to avoid introducing breaking changes, I've opted to duplicate the `Config` structure, making it private and transforming `ChatID` and `MessageThreadID` to the type `any`. Afterwards, the configuration value gets parsed and transformed back to a string, ensuring both `string` and `int` variants work.
 

# Other implementations explored

I've also explored two other options:

1. Rollback the changes introduced at grafana/grafana#63085. I've discarted this option since it was introduced for a reason and would also introduce breaking changes for all the users that now are using that featue.
2. Change the `ChatID` and `MessageThreadID` from `string` to `int`. I've also discarted this option, since it would introduce breaking changes for all the users that are already configuring Telegram using a `string` in the provisioning templates.

That's why I've opted for duplicating the struct, which is ugly but gets the work done and, more importantly, doesn't introduce any breaking change.

Let me know what you think! This issue is forcing us to stay at Grafana 9.3, which has some security vulnerabilities and not the latest features.

Thank you,

Néfix Estrada
IsardVDI

### DIFF
--- a/receivers/telegram/config.go
+++ b/receivers/telegram/config.go
@@ -28,34 +28,81 @@ type Config struct {
 	DisableNotifications  bool   `json:"disable_notifications,omitempty" yaml:"disable_notifications,omitempty"`
 }
 
+type unmarshalConfig struct {
+	BotToken              string `json:"bottoken,omitempty" yaml:"bottoken,omitempty"`
+	ChatID                any    `json:"chatid,omitempty" yaml:"chatid,omitempty"`
+	MessageThreadID       any    `json:"message_thread_id,omitempty" yaml:"message_thread_id,omitempty"`
+	Message               string `json:"message,omitempty" yaml:"message,omitempty"`
+	ParseMode             string `json:"parse_mode,omitempty" yaml:"parse_mode,omitempty"`
+	DisableWebPagePreview bool   `json:"disable_web_page_preview,omitempty" yaml:"disable_web_page_preview,omitempty"`
+	ProtectContent        bool   `json:"protect_content,omitempty" yaml:"protect_content,omitempty"`
+	DisableNotifications  bool   `json:"disable_notifications,omitempty" yaml:"disable_notifications,omitempty"`
+}
+
 func NewConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Config, error) {
 	settings := Config{}
-	err := json.Unmarshal(jsonData, &settings)
+
+	unmarshaledConfig := unmarshalConfig{}
+	err := json.Unmarshal(jsonData, &unmarshaledConfig)
 	if err != nil {
 		return settings, fmt.Errorf("failed to unmarshal settings: %w", err)
 	}
+
+	settings.BotToken = unmarshaledConfig.BotToken
+	settings.Message = unmarshaledConfig.Message
+	settings.ParseMode = unmarshaledConfig.ParseMode
+	settings.DisableWebPagePreview = unmarshaledConfig.DisableWebPagePreview
+	settings.ProtectContent = unmarshaledConfig.ProtectContent
+	settings.DisableNotifications = unmarshaledConfig.DisableNotifications
+
 	settings.BotToken = decryptFn("bottoken", settings.BotToken)
 	if settings.BotToken == "" {
 		return settings, errors.New("could not find Bot Token in settings")
 	}
-	if settings.ChatID == "" {
+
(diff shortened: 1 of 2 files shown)


NefixEstrada:
Is there any way you could look into this? @armandgrillet

jzakirov:
Hey @yuri-tceretian @santihernandezc 
Is there any blockers to merging this fix? Telegram threads not working in Grafana Alerts for 2+ years. It's a straightforward and small code change to improve quality of life.  Currently we have to use multiple chats for grouping alerts into "topics", while waiting for PR to be merged. 

Thanks in advance!

yuri-tceretian:
As is it's not possible. There's been a serious refactoring in this package. I will rebase it to main

yuri-tceretian:
The change does make sense but the copy is not necessary. It could be done much simpler using OptionalNumber field. This approach is implemented in other integrations

yuri-tceretian:
Thanks for contribution. Based on how old this PR is decided to open a new PR and incorporate my comment into it. I will make sure commit will contain @NefixEstrada as co-author.

jzakirov:
Thanks Yuri! Appreciate the response and fix 👍🏼

PR #200, opened 2024-06-05
Invert dependency between definitions and notify packages

alexweav:
Definitions in this repo, just like definitions in grafana, needs to have the smallest possible set of dependencies. This is because go-swagger traverses every package in its dependency list, and can sometimes pick up things that shouldn't be included.

The usual pattern taken from grafana/grafana is to put type conversion code in the package needing to convert from/to definitions, that way definitions doesn't depend on random other packages.

### DIFF
--- a/notify/compat.go
+++ b/notify/compat.go
@@ -1,0 +1,28 @@
+package notify
+
+import (
+	"encoding/json"
+
+	"github.com/grafana/alerting/definition"
+)
+
+func PostableAPIReceiverToAPIReceiver(r *definition.PostableApiReceiver) *APIReceiver {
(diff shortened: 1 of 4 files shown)


notify/compat.go
```diff
@@ -0,0 +1,28 @@
+package notify
+
+import (
+	"encoding/json"
+
+	"github.com/grafana/alerting/definition"
+)
+
+func PostableAPIReceiverToAPIReceiver(r *definition.PostableApiReceiver) *APIReceiver {
```
alexweav, line 9: This is just moving this function and its test to notify package. No substantial changes

yuri-tceretian (approved):

PR #173, opened 2024-03-20
Add a receiver for Amazon SNS

robbierolin:
This PR adds a receiver that will send notifications to Amazon SNS. Fixes https://github.com/grafana/alerting/issues/172

I have a corresponding change in the grafana code locally to add this receiver to the available channels that can be selected for contact points. I am guessing the process is to PR that one after this change gets merged.

Here are some example screenshots

Screenshot 2024-03-20 at 5 35 49 PM

I tested by subscribing my email to the sns topic. And verifying that I receive the emails when I send test requests.

Screenshot 2024-03-20 at 5 36 51 PM

Let me know if there's any other changes I should add or testing I should show.

PR #201, opened 2024-06-05
Update AWS SNS integration to not use grafana\aws-sdk-go

yuri-tceretian:
The integration was introduced in https://github.com/grafana/alerting/pull/173.
It imported a new module to alerting - `github.com/grafana/grafana-aws-sdk`. That module has quite a long list of dependencies, which makes all modules that use alerting pull all its dependencies. 

As long as the alerting module is vendored in Mimir project, the introduction of the Grafana's AWS SDK is sub-optmial. The alerting team decided to avoid unnecessary imports and update the integration to not use the module. 

This PR updates SNS integration with the logic copied from the upstream implementation. There is only one notable difference - because Grafana integrations use abstract webhook client, the SNS client is set up to use default HTTP client. The upstream version uses client configured for the integration https://github.com/prometheus/alertmanager/blob/33e680a34dc2e54508b9fcdb2b8c0ee81fd61d87/notify/sns/sns.go#L139C15-L139C80

cc @robbierolin


robbierolin:
The grafana-aws-sdk has logic for assuming roles in opt-in regions that I don't think would be supported with this implementation
https://github.com/grafana/grafana-aws-sdk/blob/main/pkg/awsds/sessions.go#L169
And I think some other settings that make things consistent across aws data sources - https://github.com/grafana/grafana-aws-sdk/blob/main/pkg/awsds/authSettings.go#L43

yuri-tceretian:
@robbierolin thanks for the feedback. Our main course is to unify Grafana and Prometheus integrations, and eventually get rid of the former wherever it's possible. I think going with Prometheus' implementation in this case justifies some regression in the supported feature set.

receivers/sns/config_test.go
```diff
@@ -76,9 +91,13 @@ func TestNewConfig(t *testing.T) {
 			settings: `{
 				"topic_arn": "arn:aws:sns:region:0123456789:SNSTopicName"
 			}`,
-			expectedAuthType: awsds.AuthTypeDefault.String(),
-			expectedSubject:  templates.DefaultMessageTitleEmbed,
-			expectedMessage:  templates.DefaultMessageEmbed,
+			expected: Config{
+				APIUrl:   "https://sns..amazonaws.com",
```
stevesg: Typo or intentional?

receivers/sns/config_test.go
```diff
@@ -87,16 +106,33 @@ func TestNewConfig(t *testing.T) {
 				"subject": "subject",
 				"message": "message"
 			}`,
-			expectedSubject:  "subject",
-			expectedMessage:  "message",
-			expectedAuthType: awsds.AuthTypeDefault.String(),
+			expected: Config{
+				APIUrl:   "https://sns..amazonaws.com",
```
stevesg: Typo or intentional?

receivers/sns/sns.go
```diff
@@ -21,61 +23,58 @@ import (
...
+
+	// check template error after we use them
+	if tmplErr != nil {
+		s.log.Warn("failed to template message", "error", tmplErr.Error())
 	}
 
-	snsClient := sns.New(session, aws.NewConfig().WithEndpoint(*aws.String(s.settings.APIUrl)))
 	publishOutput, err := snsClient.Publish(publishInput)
 	if err != nil {
 		s.log.Error("Failed to publish to Amazon SNS. ", "error", err)
```
stevesg, line 66: We want to return the error here. Should we check the error for whether we should retry or not, like upstream does?

```suggestion
		s.log.Error("Failed to publish to Amazon SNS. ", "error", err)
		return true, err
```

receivers/sns/sns.go
```diff
@@ -21,61 +23,58 @@ import (
...
 }
 
-func (s *Notifier) createPublishInput(ctx context.Context, as ...*types.Alert) (*sns.PublishInput, error) {
-	var err error
-	tmpl, _ := templates.TmplText(ctx, s.tmpl, as, s.log, &err)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create template: %w", err)
-	}
-
+func (s *Notifier) createPublishInput(ctx context.Context, tmpl func(string) string) (*sns.PublishInput, error) {
```
stevesg: [mega-nitpick] If this function was under `createSNSClient`, the code would be easier to compare to upstream. But this is a total nit, feel free to ignore.

stevesg (commented):
LGTM except one place I think we need a `return`. Reviewed mainly by comparing to upstream code; not tested.

receivers/sns/config_test.go
yuri-tceretian: That's the logic of the integration.

yuri-tceretian: I will change it to match the upstream

stevesg (approved):

PR #207, opened 2024-06-18
Implement default email sender

santihernandezc:
### Description

This PR adds a default email sender that implements the [EmailSender](https://github.com/grafana/alerting/blob/835aff588914c458cc1709e838d1d27d8e5ca196/receivers/email.go#L24-L26) interface.
It also includes a function that returns a factory function that can be passed to `BuildReceiverIntegrations` when creating Grafana receivers.

This can be smoke-tested using [this Mimir PoC](https://github.com/grafana/mimir/pull/8376).


receivers/email_sender.go
```diff
@@ -0,0 +1,242 @@
...
+	if data == nil {
+		data = make(map[string]any, 10)
+	}
+
+	s.setDefaultTemplateData(data)
+
+	var buffer bytes.Buffer
+	if err := s.tmpl.ExecuteTemplate(&buffer, cmd.Template, data); err != nil {
+		return nil, err
+	}
```
stevesg: The Grafana code here supports providing multiple alternative content types, but we're only doing HTML here. Should we retain text/plan support too?

receivers/email_sender.go
```diff
@@ -0,0 +1,242 @@
...
+		Body:          buffer.String(),
+		EmbeddedFiles: cmd.EmbeddedFiles,
+		ReplyTo:       cmd.ReplyTo,
+		SingleEmail:   cmd.SingleEmail,
+	}, nil
+}
+
+func (s *defaultEmailSender) setDefaultTemplateData(data map[string]any) {
+	data["AppUrl"] = s.cfg.ExternalURL
+	data["BuildVersion"] = s.cfg.Version
```
stevesg, line 132: We don't have `BuildStamp`, `EmailCodeValidHours` and `Name` here, but they're not used in our template so that makes sense.

receivers/email_sender.go
```diff
@@ -0,0 +1,242 @@
...
+	if s.cfg.CertFile != "" {
+		cert, err := tls.LoadX509KeyPair(s.cfg.CertFile, s.cfg.KeyFile)
+		if err != nil {
+			return nil, fmt.Errorf("could not load cert or key file: %w", err)
+		}
+		tlsconfig.Certificates = []tls.Certificate{cert}
+	}
+
+	d := gomail.NewDialer(host, iPort, s.cfg.AuthUser, s.cfg.AuthPassword)
+	d.TLSConfig = tlsconfig
```
stevesg, line 187: We're not setting `StartTLSPolicy` here, should we?

stevesg (approved):
I've reviewed this primarily by comparing to the Grafana code which it same from, looks good. I have left a few questions.

I'd like us to raise a follow-up issue about adding back tracing into the sender, though we can do this later on after integration into Mimir.

stevesg (approved):
Thanks, LGTM!

PR #210, opened 2024-06-20
Update TestReceivers function

fayzal-g:
This PR refactors `func (am *GrafanaAlertmanager)` to extract a lot of the functionality out into a standalone `TestReceivers` function so that it can be called from Mimir.

There is also some slight refactoring around templates, the `template` member within `GrafanaAlertmanager` has been removed as it is no longer required. Also, `templateFromContent` has been turned into a standalone helper function to no longer rely on any state, so that it can be used in multiple places.

### DIFF
--- a/notify/templates.go
+++ b/notify/templates.go
@@ -134,6 +134,8 @@ func (am *GrafanaAlertmanager) TestTemplate(ctx context.Context, c TestTemplates
 }
 
 func (am *GrafanaAlertmanager) GetTemplate() (*template.Template, error) {
+	am.reloadConfigMtx.RLock()
+
 	seen := make(map[string]struct{})
 	tmpls := make([]string, 0, len(am.templates))
 	for _, tc := range am.templates {
@@ -144,6 +146,8 @@ func (am *GrafanaAlertmanager) GetTemplate() (*template.Template, error) {
 		tmpls = append(tmpls, tc.Template)
 		seen[tc.Name] = struct{}{}
 	}
+
+	am.reloadConfigMtx.RUnlock()
 
 	tmpl, err := templateFromContent(tmpls, am.ExternalURL())
 	if err != nil {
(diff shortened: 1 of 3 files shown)


notify/templates.go
```diff
@@ -131,6 +133,26 @@ func (am *GrafanaAlertmanager) TestTemplate(ctx context.Context, c TestTemplates
 	return &results, nil
 }
 
+func (am *GrafanaAlertmanager) GetTemplate() (*template.Template, error) {
+	seen := make(map[string]struct{})
+	tmpls := make([]string, 0, len(am.templates))
```
stevesg, line 140: I think we need to hold `reloadConfigMtx` when accessing `am.templates`, because it is changed on `ApplyConfig`? Though I see in other parts of the code ( we haven't done this, that's probably a bug.

https://github.com/grafana/alerting/blob/0ddd6d9470b0f340690a597de58bbd10eaa77248/notify/templates.go#L78-L79

notify/templates.go
```diff
@@ -131,6 +133,26 @@ func (am *GrafanaAlertmanager) TestTemplate(ctx context.Context, c TestTemplates
...
+	for _, tc := range am.templates {
+		if _, ok := seen[tc.Name]; ok {
+			level.Warn(am.logger).Log("msg", "template with same name is defined multiple times, skipping...", "template_name", tc.Name)
+			continue
+		}
+		tmpls = append(tmpls, tc.Template)
+		seen[tc.Name] = struct{}{}
+	}
```
stevesg, line 148: I can't see where we did this before, does it matter if we have templates with duplicate names? Wondering if this will break for people that do have duplicates.

stevesg (commented):
Looks very good, couple comments.

notify/templates.go
fayzal-g: We currently do it inside `ApplyConfig` so I don't think it should affect any users

notify/templates.go
fayzal-g: Yeah, it does seem to me like we need it elsewhere. I've updated the code here to hold the mutex.

stevesg (approved):

PR #211, opened 2024-06-25
Bump Alertmanager version to include silence fixes

grobinson-grafana:
This commit updates Alertmanager to the include the fixes to silences in https://github.com/grafana/prometheus-alertmanager/pull/88.

`Upsert` no longer returns the id of the upserted silence as it is redundant. The id is stored in the silence and can be accessed via sil.Id. This brings the behaviour of `Upsert` inline with `Set`, which also removed the id from its return values.


yuri-tceretian (approved):
LGTM

PR #214, opened 2024-07-03
Update TestTemplate function

fayzal-g:
This PR refactors `func (am *GrafanaAlertmanager) TestTemplate` to extract a lot of the functionality out into a standalone `TestTemplate` function so that it can be called from Mimir.

### DIFF
--- a/notify/templates.go
+++ b/notify/templates.go
@@ -1,17 +1,13 @@
 package notify
 
 import (
-	"bytes"
 	"context"
-	tmplhtml "html/template"
 	"net/url"
 	tmpltext "text/template"
 
 	"github.com/go-kit/log/level"
 	"github.com/grafana/alerting/templates"
-	"github.com/prometheus/alertmanager/notify"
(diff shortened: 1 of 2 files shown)


notify/templates.go
```diff
@@ -65,72 +61,11 @@ const (
...
-			})
-		} else {
-			results.Results = append(results.Results, TestTemplatesResult{
-				Name: def,
-				Text: buf.String(),
-			})
-		}
-	}
+	am.reloadConfigMtx.RLock()
+	tmpls := am.templates
```
santihernandezc: This will use the same backing array for both slices, so any change in `tmpls` will affect `am.templates`. Using `copy(dst, src)` creates a copy of the backing array.

```suggestion
	tmpls := make([]templates.TemplateDefinition, len(am.templates))
	copy(tmpls, am.templates)
```

Note that both slices should have the same length.

santihernandezc (approved):
LGTM, please check my comment.

PR #221, opened 2024-07-10
Update Microsoft Teams to handle workflow responses differently

yuri-tceretian:
MS Teams offers multiple ways of sending notifications:
- Office 365 Incoming Webhook ([deprecated](https://devblogs.microsoft.com/microsoft365dev/retirement-of-office-365-connectors-within-microsoft-teams/), EOL on August 15, 2024)
- Workflows [link](https://make.preview.powerautomate.com/galleries/public/templates/d271a6f01c2545a28348d8f2cddf4c8f/post-to-a-channel-when-a-webhook-request-is-received) and [link](https://make.preview.powerautomate.com/galleries/public/templates/34ac6c11796143e1807d6e16c4c28050/post-to-a-chat-when-a-webhook-request-is-received)

Currently, the MS teams integration supports sending to both because it uses Adaptive Cards payload but handles response only from the Office Webhook ([link](https://learn.microsoft.com/en-us/microsoftteams/platform/webhooks-and-connectors/how-to/connectors-using?tabs=cURL%2Ctext1#send-messages-using-curl-and-powershell)), and treats all responses that do not have character `1` in the response as failed requests. 

The workflow webhook does not follow this contract, and it causes successful requests to be considered as failed.

This PR fixes the response handling. It proposes different logic for handling responses that is applied based on the URL. If the URL host ends with ".webhook.office.com, " the request is considered an Incoming Webhook. 
I think this is a safe assumption because in the case of a false-negative match, the status is calculated by status code.

Related to: https://github.com/grafana/grafana/issues/90139

### DIFF
--- a/receivers/teams/models.go
+++ b/receivers/teams/models.go
@@ -1,0 +1,10 @@
+package teams
+
+type errorResponse struct {
+	Error errorBody `json:"error"`
+}
+
+type errorBody struct {
+	Code    string `json:"code"`
+	Message string `json:"message"`
+}
--- a/receivers/teams/teams.go
+++ b/receivers/teams/teams.go
@@ -316,6 +316,7 @@ func (tn *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error
 	if err != nil {
 		return false, fmt.Errorf("failed to parse URL: %w", err)
 	}
+	// TODO: remove it after August 15. Office webhooks are deprecated and are removed on August 15, https://devblogs.microsoft.com/microsoft365dev/retirement-of-office-365-connectors-within-microsoft-teams/
 	if strings.HasSuffix(parsed.Host, "webhook.office.com") {
 		// Teams sometimes does not use status codes to show when a request has failed. Instead, the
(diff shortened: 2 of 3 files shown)


santihernandezc:
LGTM 👍 
I tested this by using this branch within `grafana/grafana`. Notifications are successfully sent via workflows, there's no error message in the logs, and the notification attempt is recorded in the contact point metadata.
Testing the contact point works too, the toast with the error message we're getting in `main` is not there anymore.

santihernandezc (approved):
LGTM 👍
I tested this by using this branch within grafana/grafana. Notifications are successfully sent via workflows, there's no error message in the logs, and the notification attempt is recorded in the contact point metadata.
Testing the contact point works too, the toast with the error message we're getting in main is not there anymore.

champak:
@yuri-tceretian thanks very much for this fix! I had a minor clarification regarding the TO-DO item [6dd3b5a](https://github.com/grafana/alerting/pull/221/commits/6dd3b5a48ea731984b1e15355262533a1d890fe8). The comment says " // TODO: remove it after August 15. Office webhooks are deprecated and are removed on August 15 .....". From Microsoft's blog it sounds like the webhooks are actually deprecated on 10/1 and August 15 is the date beyond which no new instances of these webhooks shall be allowed. If that is correct, would it make sense to leave the code around till after 10/1? It might make transitioning simpler if both modes are working for a bit longer. In any case hoping to see these changes in a release soon!

PR #228, opened 2024-08-07
Alerting: Update Error to string in TestIntegrationConfigResult and TestTemplatesErrorResult

fayzal-g:
Also adds JSON tags for `TestReceiversResult`, `TestReceiverResult` and `TestIntegrationConfigResult` so they are lower case


stevesg (approved):
LGTM

Talked with Fayzal about this offline, we debated whether to add an additional "error type" next to the error, or have the testing code generate the response code. We decided the latter was marginally better.

PR #227, opened 2024-07-30
Add MQTT receiver

alexander-akhmetov:
Adds an MQTT (v3.1) receiver.

Fixes https://github.com/grafana/grafana/issues/16858

* PR to Grafana: https://github.com/grafana/grafana/pull/91487
* PR to terraform provider: https://github.com/grafana/terraform-provider-grafana/pull/1746

<details>

<summary>How to start an MQTT broker for testing</summary>

[mosquitto](https://github.com/eclipse/mosquitto/tree/master) can be used as a local MQTT broker for testing. Create the config file ([full config file reference](https://github.com/eclipse/mosquitto/blob/master/mosquitto.conf)):

```
$ cat mosquitto.conf
listener 1883 0.0.0.0
allow_anonymous true  # set to false to disallow anonymous users
```

Start the server:
```
docker run -it -p 127.0.0.1:1883:1883 -v ./mosquitto.conf:/mosquitto/config/mosquitto.conf eclipse-mosquitto
```

#### usernames and passwords

Install mosquitto for `mosquitto_passwd` tool. Create the passwords file:

```
mosquitto_passwd -c passwords.txt user1
```

Add the following to the `mosquitto.conf`:

```
password_file /mosquitto/config/passwords.txt
```

Start the server:
```
docker run -it -p 127.0.0.1:1883:1883 -v ./mosquitto.conf:/mosquitto/config/mosquitto.conf -v ./passwords.txt:/mosquitto/config/passwords.txt eclipse-mosquitto
```

</details>

**Screenshots**

image

image

PR #232, opened 2024-08-22
Add TLS, QoS and retain options to the MQTT receiver

alexander-akhmetov:
Add TLS, QoS and Retain parameters to the MQTT receiver.

Grafana PR: https://github.com/grafana/grafana/pull/92331

Related: https://github.com/grafana/grafana/issues/16858 and https://github.com/grafana/alerting/pull/227

### DIFF
--- a/receivers/mqtt/config.go
+++ b/receivers/mqtt/config.go
@@ -25,7 +25,7 @@ type Config struct {
 	Password      string                   `json:"password,omitempty" yaml:"password,omitempty"`
 	QoS           receivers.OptionalNumber `json:"qos,omitempty" yaml:"qos,omitempty"`
 	Retain        bool                     `json:"retain,omitempty" yaml:"retain,omitempty"`
-	TLS           *receivers.TLSConfig     `json:"tls,omitempty" yaml:"tls,omitempty"`
+	TLSConfig     *receivers.TLSConfig     `json:"tlsConfig,omitempty" yaml:"tlsConfig,omitempty"`
 }
 
 func NewConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Config, error) {
@@ -68,10 +68,10 @@ func NewConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Confi
 
 	settings.Password = decryptFn("password", settings.Password)
 
-	if settings.TLS != nil {
-		settings.TLS.CACertificate = decryptFn("tlsCACertificate", settings.TLS.CACertificate)
-		settings.TLS.ClientCertificate = decryptFn("tlsClientCertificate", settings.TLS.ClientCertificate)
-		settings.TLS.ClientKey = decryptFn("tlsClientKey", settings.TLS.ClientKey)
+	if settings.TLSConfig != nil {
+		settings.TLSConfig.CACertificate = decryptFn("tlsCACertificate", settings.TLSConfig.CACertificate)
+		settings.TLSConfig.ClientCertificate = decryptFn("tlsClientCertificate", settings.TLSConfig.ClientCertificate)
+		settings.TLSConfig.ClientKey = decryptFn("tlsClientKey", settings.TLSConfig.ClientKey)
 	}
 
 	return settings, nil
--- a/receivers/mqtt/config.go
+++ b/receivers/mqtt/config.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
+	"net/url"
 
 	"github.com/grafana/alerting/receivers"
 	"github.com/grafana/alerting/templates"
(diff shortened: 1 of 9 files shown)


receivers/mqtt/config.go
```diff
@@ -16,14 +16,19 @@ const (
...
-	Password           string `json:"password,omitempty" yaml:"password,omitempty"`
-	InsecureSkipVerify bool   `json:"insecureSkipVerify,omitempty" yaml:"insecureSkipVerify,omitempty"`
+	BrokerURL            string                   `json:"brokerUrl,omitempty" yaml:"brokerUrl,omitempty"`
+	ClientID             string                   `json:"clientId,omitempty" yaml:"clientId,omitempty"`
+	Topic                string                   `json:"topic,omitempty" yaml:"topic,omitempty"`
+	Message              string                   `json:"message,omitempty" yaml:"message,omitempty"`
+	MessageFormat        string                   `json:"messageFormat,omitempty" yaml:"messageFormat,omitempty"`
+	Username             string                   `json:"username,omitempty" yaml:"username,omitempty"`
+	Password             string                   `json:"password,omitempty" yaml:"password,omitempty"`
+	QoS                  receivers.OptionalNumber `json:"qos,omitempty" yaml:"qos,omitempty"`
```
yuri-tceretian: Let's use int. IIRC OptionalNumber was introduced to support backward compatibility with integrations from legacy alerting.

receivers/mqtt/config.go
```diff
@@ -16,14 +16,19 @@ const (
...
+	InsecureSkipVerify   bool                     `json:"insecureSkipVerify,omitempty" yaml:"insecureSkipVerify,omitempty"`
+	Retain               bool                     `json:"retain,omitempty" yaml:"retain,omitempty"`
+	TLSCACertificate     string                   `json:"tlsCACertificate,omitempty" yaml:"tlsCACertificate,omitempty"`
+	TLSClientCertificate string                   `json:"tlsClientCertificate,omitempty" yaml:"tlsClientCertificate,omitempty"`
+	TLSClientKey         string                   `json:"tlsClientKey,omitempty" yaml:"tlsClientKey,omitempty"`
```
yuri-tceretian: It would be great if we group these fields into a struct (similar to what upstream does). This schema will let us group fields in UI as well.
There are some problems with nested secrets right now but it should be fixed in https://github.com/grafana/grafana/pull/92035.

yuri-tceretian: Later we may reuse it in other integrations such as webhook

receivers/mqtt/config.go
```diff
@@ -16,14 +16,16 @@ const (
...
+	BrokerURL     string                   `json:"brokerUrl,omitempty" yaml:"brokerUrl,omitempty"`
+	ClientID      string                   `json:"clientId,omitempty" yaml:"clientId,omitempty"`
+	Topic         string                   `json:"topic,omitempty" yaml:"topic,omitempty"`
+	Message       string                   `json:"message,omitempty" yaml:"message,omitempty"`
+	MessageFormat string                   `json:"messageFormat,omitempty" yaml:"messageFormat,omitempty"`
+	Username      string                   `json:"username,omitempty" yaml:"username,omitempty"`
+	Password      string                   `json:"password,omitempty" yaml:"password,omitempty"`
+	QoS           receivers.OptionalNumber `json:"qos,omitempty" yaml:"qos,omitempty"`
+	Retain        bool                     `json:"retain,omitempty" yaml:"retain,omitempty"`
+	TLS           *receivers.TLSConfig     `json:"tls,omitempty" yaml:"tls,omitempty"`
```
yuri-tceretian: Let's call it `TLSConfig` and use json name `tls_config`, the same as in https://github.com/grafana/prometheus-alertmanager/blob/3d49ff83c74e253fc60b7659d952b24c59c44962/config/notifiers.go#L264

yuri-tceretian (approved):
LGTM

receivers/mqtt/mqtt.go
```diff
@@ -102,6 +113,39 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 	return true, nil
 }
 
+func (n *Notifier) buildTLSConfig() (*tls.Config, error) {
```
titolins: Since there's a new `TLSConfig` added to `receivers` package, why not generalizing this also? I mean, making it a function in the `receivers` package instead that takes `TLSConfig` and returns `tls.Config`.

It would also be similar to how [prometheus](https://github.com/prometheus/common/blob/main/config/http_config.go#L1025) handles it

titolins (commented):
Hey @alexander-akhmetov, hope I'm on time for a last comment here. I'm investigating how to inject TLS config in the webhook receiver and this could help a lot (in the scope of https://github.com/grafana/grafana/issues/9548).

I could also tackle the suggested changes in a separate PR once you merge it, if you'd prefer 👍

receivers/mqtt/mqtt.go
alexander-akhmetov: Makes sense! I changed it, but I find the naming a bit confusing 😅 Maybe `ToCryptoTLSConfig` or something else would be better, what do you think?

https://github.com/grafana/alerting/commit/99614a5e12945fe18a9b44ff64751b1658ef1837

titolins: Nice, thanks! Yeah, agreed about the naming - it can be a bit confusing. `ToCryptoTLSConfig` sounds better 👍

titolins (approved):
🚀

receivers/mqtt/mqtt.go
alexander-akhmetov: Updated :)

PR #234, opened 2024-08-23
[Enhancement] Make `APIURL` for Pushover configurable

mrusme:
I'm proposing an enhancement to the Pushover receiver, namely the ability to configure the `APIURL`. My motivation for that is that I'm using [an open-source alternative to Pushover](https://github.com/mrusme/overpush) which mimics its `messages.json` endpoint and is hence a drop-in replacement for Pushover.

I would like to be able to configure Grafana to use it as an alert receiver. The only thing that is missing in order to do so, is the ability to configure the `APIURL`.

Kindly consider this enhancement. If approved, I'm willing to go ahead and PR the required changes in the other Grafana components for this to work. Thank you.

### DIFF
--- a/receivers/pushover/config.go
+++ b/receivers/pushover/config.go
@@ -10,6 +10,7 @@ import (
 )
 
 type Config struct {
+	APIURL           string
 	UserKey          string
 	APIToken         string
 	AlertingPriority int64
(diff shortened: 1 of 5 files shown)


mrusme:
I have built a custom solution to this issue into [my own Pushover implementation](https://github.com/mrusme/overpush?tab=readme-ov-file#grafana) and will hence close this PR. It's sad that this PR didn't receive any attention at all, though, as allowing users to change the Pushover URL would have been an easier solution.

PR #236, opened 2024-08-28
MQTT receiver: Add a setting to append group key to the topic name

alexander-akhmetov:
Adds a new setting, `AddGroupKeyToTopic`. If true, appends the group key to the topic name.

Grafana PR: https://github.com/grafana/grafana/pull/92618

Related: https://github.com/grafana/grafana/issues/16858 and https://github.com/grafana/alerting/pull/227


yuri-tceretian (dismissed):
LGTM

receivers/mqtt/mqtt.go
```diff
@@ -86,10 +87,20 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
...
 	}()
 
+	topic := n.settings.Topic
+	if n.settings.AddGroupKeyToTopic {
+		groupKey, err := notify.ExtractGroupKey(ctx)
+		if err != nil {
+			n.log.Error("Failed to extract group key", "error", err.Error())
+			return false, fmt.Errorf("Failed to extract group key: %s", err.Error())
+		}
+		topic = path.Join(n.settings.Topic, string(groupKey))
```
yuri-tceretian, line 108: there are couple of edge cases that I think we should take care of:
1. Let's trim `/` at the ends, and replace all occurrences with `-` or `_`. 
2. Same about wildcards: `+` and `#`
3. MQTT puts restriction on length of the topic, up to 65535. Perhaps it makes sense to check it

yuri-tceretian: the format of group key is like `{parent_matchers[/child_matchers]+}:{group_labels}` . I wonder what you think if we unwrap the matchers brackets and the topic will resemble the routing tree like.
`.../parent_matchers/child_matcher/child_matcher/group_key`

PR #239, opened 2024-08-30
Ensure ActiveTimeIntervals are propagated similar to MuteTimeIntervals

JacobsonMT:
Currently `ActiveTimeIntervals` are not validated or propagated in the same way as `MuteTimeIntervals`. This fixes that discrepancy and improves testing around complex route compats.

Followup to https://github.com/grafana/alerting/pull/229
Related to https://github.com/grafana/mimir/issues/8954

### DIFF
--- a/definition/alertmanager_validation.go
+++ b/definition/alertmanager_validation.go
@@ -85,6 +85,8 @@ func (r *Route) ValidateReceivers(receivers map[string]struct{}) error {
 	return nil
 }
 
+// ValidateMuteTimes validates that all mute time intervals referenced by the route exist.
+// TODO: Can be removed once grafana/grafan uses ValidateTimeIntervals instead.
 func (r *Route) ValidateMuteTimes(timeIntervals map[string]struct{}) error {
 	for _, name := range r.MuteTimeIntervals {
 		if _, exists := timeIntervals[name]; !exists {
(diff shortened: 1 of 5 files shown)


definition/alertmanager_validation.go
```diff
@@ -82,14 +85,29 @@ func (r *Route) ValidateReceivers(receivers map[string]struct{}) error {
...
-		err := child.ValidateMuteTimes(muteTimes)
+		err := child.ValidateMuteTimes(timeIntervals)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *Route) ValidateActiveTimes(timeIntervals map[string]struct{}) error {
```
santihernandezc: We're not using this method anywhere.
We could change `ValidateMuteTimes -> ValidateTimeIntervals` and check both `MuteTimeIntervals` and `ActiveTimeIntervals` in the same place.

santihernandezc (approved):
LGTM! Just one small thing that should be addressed before merging.
I don't need to review this again.

definition/alertmanager_validation.go
JacobsonMT: We use it in [grafana/grafana](https://github.com/grafana/grafana/blob/7b919e32771b8228684234d5cc7262d915a96534/pkg/services/ngalert/provisioning/notification_policies.go#L83), but I agree one method would be better. I'll add it as a 2nd method for now so we don't needlessly break grafana/grafana compilation on dependency update, then we can remove at will.

santihernandezc: Ah ok! The plan is to use `ValidateActiveTimes` in grafana/grafana right next to the call to `ValidateMuteTimes` then. Sounds good 🙂

PR #240, opened 2024-08-31
Include screenshot URL to Telegram message

### LINKED ISSUE #123: Include screenshot URL to Telegram message
Currently, Telegram integration sends two messages when an alert has a screenshot attached to it. 
However, Grafana screenshot service can upload screenshots to a remote host and provides URL to the image in the alert. 

When Grafana is configured this way, we probably can let users control how screenshots are sent by Telegram integration: add URL to the first message or send it as the second message.

nutmos:
This PR is for adding screenshot URL to Telegram messages.

This PR introduces `IncludeScreenshotURL` for a Telegram channel. When switch it on, all images will be sent to Telegram into 1 message as URLs. When switch it off, an image will be sent with the separate message.

Fixes #123

### DIFF
--- a/receivers/telegram/telegram.go
+++ b/receivers/telegram/telegram.go
@@ -119,7 +119,16 @@ func (tn *Notifier) buildTelegramMessage(ctx context.Context, as []*types.Alert)
 
 	tmpl, _ := templates.TmplText(ctx, tn.tmpl, as, tn.log, &tmplErr)
 	// Telegram supports 4096 chars max
-	messageText, truncated := receivers.TruncateInRunes(tmpl(tn.settings.Message), telegramMaxMessageLenRunes)
+	rawMessage := tmpl(tn.settings.Message)
+	if tn.settings.IncludeScreenshotURL {
+		_ = images.WithStoredImages(ctx, tn.log, tn.images, func(_ int, image images.Image) error {
+			if len(image.URL) != 0 {
+				rawMessage += image.URL
+			}
+			return nil
+		}, as...)
+	}
+	messageText, truncated := receivers.TruncateInRunes(rawMessage, telegramMaxMessageLenRunes)
 	if truncated {
 		key, err := notify.ExtractGroupKey(ctx)
 		if err != nil {
--- a/receivers/telegram/telegram.go
+++ b/receivers/telegram/telegram.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"mime/multipart"
+	"net/url"
 	"os"
 
 	"github.com/prometheus/alertmanager/notify"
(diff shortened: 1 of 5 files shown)


nutmos:
@yuri-tceretian Could you please help me review the PR? Thank you.

yuri-tceretian:
Hi @nutmos, thank you for contributing to the project! I really appreciate it. While I am reviewing the changes, can you please add more information about what the pull request does, format of the notification before and after the change?

nutmos:
@yuri-tceretian I updated the description already. Thanks for your feedback.

receivers/telegram/telegram.go
```diff
@@ -74,34 +74,37 @@ func (tn *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error
...
-		cmd, err = tn.newWebhookSyncCmd("sendPhoto", func(w *multipart.Writer) error {
-			f, err := os.Open(image.Path)
-			if err != nil {
-				return fmt.Errorf("failed to open image: %w", err)
-			}
-			defer func() {
-				if err := f.Close(); err != nil {
-					tn.log.Warn("failed to close image", "error", err)
+	// Works only if IncludeScreenshotURL is set to false.
+	if !tn.settings.IncludeScreenshotURL {
```
yuri-tceretian: Let's invert this expression. Go's best practices recommend reducing nesting, and this is a good candidate.

yuri-tceretian:
I gave it a closer look and unfortunately, I cannot accept the proposed solution. 

As you may know, there is a limit to the maximum length of the message - 4096 of UTF-16 characters. The URL, especially when it's a link provided by a cloud provider, can be very long (200-300 characters). Also, in the case of multiple alerts in a group, there could be multiple images for the message. Appending them to the message as text will likely make them truncated, which is no better than dropping them altogether. Also, from a perception standpoint, long URLs are not useful and clutter the message, especially when they go one after another.

I can think of an alternative solution - the request model of `sendMessage` API offers field `entities`  (see https://core.telegram.org/bots/api#sendmessage). It allows clients to implement some WYSIWYG capabilities and format the plain text message. For example 
```
curl --location 'https://api.telegram.org/bot$BOT_TOKEN/sendMessage' \
--form 'chat_id="$CHAT_ID"' \
--form 'entities="[{\"type\":\"text_link\", \"offset\": 10, \"length\": 7, \"url\":\"http://grafana.com\"}]"' \
--form 'text="Hello, 🎉 Grafana\!"'
```
produces the following message 

<details><summary>Screenshot</summary>

image

</details>

This approach will let us save a lot of space in the message, and for example append screenshots like 

I think the latter is the best solution here. However, it works only when `parse_mode` is not specified. Thus, it's not universally good solution, which is ok, I think. We can document this specifics, and, ideally, add URLs to the template data so users could add URLs to their custom templates.

yuri-tceretian (changes requested):
I gave it a closer look and unfortunately, I cannot accept the proposed solution. 

As you may know, there is a limit to the maximum length of the message - 4096 of UTF-16 characters. The URL, especially when it's a link provided by a cloud provider, can be very long (200-300 characters). Also, in the case of multiple alerts in a group, there could be multiple images for the message. Appending them to the message as text will likely make them truncated, which is no better than dropping them altogether. Also, from a perception standpoint, long URLs are not useful and clutter the message, especially when they go one after another.

I can think of an alternative solution - the request model of `sendMessage` API offers field `entities`  (see https://core.telegram.org/bots/api#sendmessage). It allows clients to implement some WYSIWYG capabilities and format the plain text message. For example 
```
curl --location 'https://api.telegram.org/bot$BOT_TOKEN/sendMessage' \
--form 'chat_id="$CHAT_ID"' \
--form 'entities="[{\"type\":\"text_link\", \"offset\": 10, \"length\": 7, \"url\":\"http://grafana.com\"}]"' \
--form 'text="Hello, 🎉 Grafana\!"'
```
produces the following message 

<details><summary>Screenshot</summary>

image

</details>

This approach will let us save a lot of space in the message, and for example append screenshots like 
```
.... Screenshots: [1],[2],[3]
```
where [1] is a clickable link.

However, there is a downside too - this approach works only when `parse_mode` is not specified. However, the integration supports different formats such as HTML or Markdown. So, can't use entities in those cases. I think this is ok if we provide some alternatives (does not have to be in this PR).
Ideally, in such a configuration, we can populate template data with URLs. There are already fields in the template data model
https://github.com/grafana/alerting/blob/e81931acc11bf01773d011f43233546d062a93fa/templates/template_data.go#L50-L51
Also, there are examples in webhook and email integrations.

Actually, I think I can accept PR that just sets the ImageURL field in template data, and leaves it up to the custom template to include the URL in the message.

nutmos:
@yuri-tceretian Thank you for your suggestions. That makes sense. I will do it later.

BloodDK:
UP!!! we need this feature!

yuri-tceretian:
I will prioritize reviewing it. Sorry for the delay

receivers/telegram/telegram.go
```diff
@@ -116,7 +123,18 @@ func (tn *Notifier) buildTelegramMessage(ctx context.Context, as []*types.Alert)
...
 	tmpl, _ := templates.TmplText(ctx, tn.tmpl, as, tn.log, &tmplErr)
 	// Telegram supports 4096 chars max
-	messageText, truncated := receivers.TruncateInRunes(tmpl(tn.settings.Message), telegramMaxMessageLenRunes)
+	rawMessage := tmpl(tn.settings.Message)
+	if tn.settings.IncludeScreenshotURL {
+		_ = images.WithStoredImages(ctx, tn.log, tn.images, func(_ int, image images.Image) error {
+			// Change to add the template ImageURL
+			imageURL, err := url.Parse(image.URL)
+			if err == nil {
+				tn.tmpl.ExternalURL = imageURL
```
yuri-tceretian, line 132: This will not work because the ExternalURL is used in other places. We need to figure out a better place for this URL. For now, I would add it to annotations under "Screenshot URL".

yuri-tceretian (changes requested):
See my comment below.

receivers/telegram/telegram.go
yuri-tceretian: I configured my local instance to store screenshot in cloud. That's what I get 
image

All urls are replaced with screenshot URL which is not correct.

hawk0774:
Коллеги, печальная ситуация с данным функционалом, раньше все одним сообщением отправлялось, теперь какой-то ужас

BloodDK:
> Коллеги, печальная ситуация с данным функционалом, раньше все одним сообщением отправлялось, теперь какой-то ужас

Тут вообще все печально, вставка url в текст и вынос всех картинок на публичный доступ, только ради того, чтобы соединить сообщение с фото.. 
Имхо, проблема в самом api телеграмма, которая принимает либо только текст через метод SendMessage
Либо фото с текстом через SendPhoto, но с нюансом короткого текста. 
Я бы мог конечно написать прокси сервис, так сказать webhook, который захватывает все сообщения при отправке алерта от графаны, + полезет в графану по panelid из текста и сделав скриншот, отправит все через метод SendPhoto. Но это тоже костыль, учитывая что в алертах много текста, это тупо не пролезет в телеграм api 😁 
Ну и вынос всех картинок в публичный доступ - это риски информ. безопасности, то что предлагает автор хороший вариант, и фото будет снизу вроде бы, но не секюрно.
А прокся между графаной и телегой, это костыль, где фото всегда будет сверху сообщения 😁  стоит ли весь этот гемор телодвижений?

yuri-tceretian:
> Коллеги, печальная ситуация с данным функционалом, раньше все одним сообщением отправлялось, теперь какой-то ужас

Если у вас есть предложения по улучшению данного функционала, мы всегда рады рассмотеть запрос на изменения (pull request). Даже задача (issue) с описанием предложения - было бы замечательно. По-моему, мое предложения по использованию форматирования https://github.com/grafana/alerting/pull/240#pullrequestreview-2385745099 довольно неплохой выход из ситуации. По-крайней мере, лучше чем сейчас. 

Я рад видеть русскоязычные коментарии, и с удовольствием пообщался бы на русском языке, но давайте уважать всех и использовать английский, коллеги :)

Issue #235, opened 2024-08-28
Slack files.upload Broken for New Slack Apps as of May 2024


ESNewmanium:
Image uploads to Slack using `ngalert.notifier.slack` are no longer working as of May 2024. [Slack API has deprecated this method](https://api.slack.com/methods/files.upload#markdown) and prevents newly created apps from using it. The Slack API responds with "method_deprecated" and the upload fails.

An example of an attempted image upload is below.

```
logger=ngalert.notifier.slack notifierUID=e41965c3-d944-4377-b3fa-816762249833 t=2024-08-28T08:24:37.84186656-05:00 level=debug msg="Looks like we are using the Slack API, have set the Bearer token for this request"
logger=ngalert.notifier.slack notifierUID=e41965c3-d944-4377-b3fa-816762249833 t=2024-08-28T08:24:38.095552788-05:00 level=debug msg="The request was successful"
logger=ngalert.notifier.slack notifierUID=e41965c3-d944-4377-b3fa-816762249833 t=2024-08-28T08:24:38.096132436-05:00 level=debug msg="Uploadimg image" image=f58a0199-76ce-4169-8e2f-f3c544211250
logger=ngalert.notifier.slack notifierUID=e41965c3-d944-4377-b3fa-816762249833 t=2024-08-28T08:24:38.096233945-05:00 level=debug msg="Sending multipart request to files.upload"
logger=ngalert.notifier.slack notifierUID=e41965c3-d944-4377-b3fa-816762249833 t=2024-08-28T08:24:38.198886857-05:00 level=error msg="The request was unsuccessful" body="{\"ok\":false,\"error\":\"method_deprecated\"}" err=method_deprecated
logger=ngalert.notifier.slack notifierUID=e41965c3-d944-4377-b3fa-816762249833 alert="Unexpected Container Restart[5a4c147][active]" t=2024-08-28T08:24:38.198943018-05:00 level=error msg="Failed to attach image to notification" error="failed to send request: failed to send request: method_deprecated"
logger=ngalert.notifier.slack notifierUID=e41965c3-d944-4377-b3fa-816762249833 t=2024-08-28T08:24:38.198961721-05:00 level=error msg="Failed to upload image" err="failed to send request: failed to send request: method_deprecated"

```

moustafab:
This is fixed in #256 and will be released in grafana v11.5.x. Thanks for filing this issue!

PR #242, opened 2024-09-02
fix: deprecated slack files.upload is replaced with new slack bot APIs

iamatulsingh:
#### ENHANCEMENT

```txt
The original web API method for uploading files to Slack, [files.upload](https://api.slack.com/methods/files.upload), is being sunset on March 11, 2025.

As of May 16, 2024, newly-created apps are no longer able to use this API method.

Existing apps & integrations should migrate away from files.upload and instead leverage a combination of two APIs: [files.getUploadURLExternal](https://api.slack.com/methods/files.getUploadURLExternal) and [files.completeUploadExternal](https://api.slack.com/methods/files.completeUploadExternal). The use of these two methods is more reliable, especially when uploading large files.

A detailed breakdown of how to use these two API methods is described in our [Uploading files documentation](https://api.slack.com/messaging/files#uploading_files).
```

Mentioned in some `issues` [#235](https://github.com/grafana/alerting/issues/235)

As mentioned here https://api.slack.com/changelog/2024-04-a-better-way-to-upload-files-is-here-to-stay I mad changes to use new APIs now with backward compatibility for `Slack`.


stevesg (commented):
Thank you for working on this!

It's not commonplace in Grafana to reach into environment variables directly, so we'll need to find an alternative to this, some options:
1. Don't make it configurable, just use the new API and be done with it.
2. Make it configurable via Grafana global config level, pass the config down
3. Make it configurable on the integration level

iamatulsingh:
> Thank you for working on this!
> 
> It's not commonplace in Grafana to reach into environment variables directly, so we'll need to find an alternative to this, some options:
> 
> 1. Don't make it configurable, just use the new API and be done with it.
> 2. Make it configurable via Grafana global config level, pass the config down
> 3. Make it configurable on the integration level

Thank you for your time. I have used the new API and removed the old one completely.

PR #247, opened 2024-09-19
add support for TLSConfig in webhook receiver

titolins:
Add TLS configuration support for webhook receivers

## Related PR's
- https://github.com/grafana/grafana/pull/93513
- https://github.com/grafana/mimir/pull/9341


receivers/webhook/webhook.go
```diff
@@ -111,13 +112,21 @@ func (wn *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error
 		return false, tmplErr
 	}
 
+	var tlsConfig *tls.Config
+	if wn.settings.TLSConfig != nil {
+		if tlsConfig, err = wn.settings.TLSConfig.ToCryptoTLSConfig(); err != nil {
```
fayzal-g, line 117: Slightly unrelated to this PR, but we are pretty much redefining the library here: https://github.com/grafana/dskit/blob/main/crypto/tls/tls.go#L87

It would need to be updated to take the certs/keys as a string though (currently only takes filepaths)

receivers/webhook/webhook_test.go
```diff
@@ -331,6 +342,105 @@ func TestNotify(t *testing.T) {
...
+			alerts: []*types.Alert{
+				{
+					Alert: model.Alert{
+						Labels:      model.LabelSet{"alertname": "alert1", "lbl1": "val1"},
+						Annotations: model.LabelSet{"ann1": "annv1", "__dashboardUid__": "abcd", "__panelId__": "efgh"},
+					},
+				},
+			},
+			expMsgError: fmt.Errorf("Unable to use the provided CA certificate"),
+		},
```
fayzal-g, line 512: Can we add another test case for `InsecureSkipVerify: true`?

titolins: Sure, done now 👍

receivers/webhook/webhook.go
titolins: yeah, that's a good point. You'd mention that library but I just went with the same setup that we had for mqtt and forgot about it.
The dskit is more complete also. I could adapt both to use that instead or create another issue to standardize it after this is merged - wdyt?

receivers/webhook/fixtures/client.key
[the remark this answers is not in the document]
titolins: this is a mock key used for unit tests, nothing to worry about (reported as false positive)

receivers/webhook/webhook_test.go
```diff
@@ -19,6 +20,16 @@ import (
 	"github.com/grafana/alerting/templates"
 )
 
+//go:embed fixtures/ca.pem
+var caCert string
+
+//
```
fayzal-g: Nit: Remove

fayzal-g (approved):
lgtm! Just one small nit

receivers/webhook/webhook_test.go
titolins: good catch, removed now 👍

PR #248, opened 2024-09-25
Fix: Check and clear `tmplErr` after executing the title template (Slack integration)

### LINKED ISSUE #80224: Alerting: Templating plus Slack breaks when name for the notification template contain square brackets
### What happened?

I use Grafana Alerting with Slack as notification platform.

To increase readability, I create several notification templates. I would like to use the following naming pattern when I add a new template:

- [Slack] Message Title
- [Slack] Message Text
- [EMail] Subject
- [EMail] Body

When I try to test a Slack contact point with the existing templates, I get an error message:

```text
logger=ngalert.notifier.slack notifierUID=*** t=*** level=error msg="The request was unsuccessful" body="{\"ok\":false,\"error\":\"invalid_arguments\",\"warning\":\"missing_charset\",\"response_metadata\":{\"messages\":[\"[ERROR] missing required field: channel\"],\"warnings\":[\"missing_charset\"]}}" err=invalid_arguments
```

If I remove the square brackets from the template names and then test the Slack contact point again, the test message is sent without an error.

In the documentation I cannot find any hint that there are character restrictions for template names.

### What did you expect to happen?

I expect that all characters are allowed in the template names. If there are any character restrictions, I expect the documentation to contain appropriate information.

### Did this work before?

I don't know

### How do we reproduce it?

See "What happened"

### Is the bug inside a dashboard panel?

_No response_

### Environment (with versions)?

Grafana: 10.1.1
OS: MacOS Sonoma
Browser: Chrome

### Grafana platform?

Kubernetes

### Datasource(s)?

_No response_

santihernandezc:
Because of [how we handle templating errors](https://github.com/grafana/alerting/blob/6c25eb6eff103307621ef51bfa90192d839e96db/templates/template_data.go#L232-L243), a bad `title` template in a Slack receiver results in no value for the `channel` field. If a title template fails to be executed and we're using a bot token, we'll get back an error from the Slack API and the notification won't be sent.

```json
{
  "ok": false,
  "error": "invalid_arguments",
  "warning": "missing_charset",
  "response_metadata": {
    "messages": [
      "[ERROR] missing required field: channel"
    ],
    "warnings": [
      "missing_charset"
    ]
  }
}
```

This PR addresses this issue by clearing the error before we continue templating other fields. It also removes the `missing_charset` warning by sending `Content-Type": "application/json; charset=utf-8`.

Fixes https://github.com/grafana/grafana/issues/80224

### DIFF
--- a/receivers/slack/slack.go
+++ b/receivers/slack/slack.go
@@ -315,6 +315,10 @@ func (sn *Notifier) createSlackMessage(ctx context.Context, alerts []*types.Aler
 		}
 		sn.log.Warn("Truncated title", "key", key, "max_runes", slackMaxTitleLenRunes)
 	}
+	if tmplErr != nil {
+		sn.log.Warn("failed to template Slack title", "error", tmplErr.Error())
+		tmplErr = nil
+	}
 
 	req := &slackMessage{
 		Channel:   tmpl(sn.settings.Recipient),
@@ -400,7 +404,7 @@ func (sn *Notifier) sendSlackMessage(ctx context.Context, m *slackMessage) (stri
 		return "", fmt.Errorf("failed to create HTTP request: %w", err)
 	}
 
-	request.Header.Set("Content-Type", "application/json")
+	request.Header.Set("Content-Type", "application/json; charset=utf-8")
 	request.Header.Set("User-Agent", "Grafana")
 	if sn.settings.Token == "" {
 		if sn.settings.URL == APIURL {
--- a/receivers/slack/slack_test.go
+++ b/receivers/slack/slack_test.go
@@ -470,6 +470,45 @@ func TestNotify_PostMessage(t *testing.T) {
 			},
 		},
 	}, {
+		name: "Errors in templates, message is sent",
+		settings: Config{
+			EndpointURL:    APIURL,
+			URL:            APIURL,
+			Token:          "1234",
+			Recipient:      "#test",
+			Text:           `{{ template "undefined" . }}`,
+			Title:          `{{ template "undefined" . }}`,
+			Username:       `{{ template "undefined" . }}`,
+			IconEmoji:      `{{ template "undefined" . }}`,
+			IconURL:        "",
+			MentionChannel: "",
+			MentionUsers:   nil,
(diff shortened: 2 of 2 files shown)


receivers/slack/slack.go
```diff
@@ -400,7 +404,7 @@ func (sn *Notifier) sendSlackMessage(ctx context.Context, m *slackMessage) (stri
 		return "", fmt.Errorf("failed to create HTTP request: %w", err)
 	}
 
-	request.Header.Set("Content-Type", "application/json")
+	request.Header.Set("Content-Type", "application/json; charset=utf-8")
```
santihernandezc, line 407: This removes a warning sent back by Slack.

receivers/slack/slack.go
```diff
@@ -315,6 +315,10 @@ func (sn *Notifier) createSlackMessage(ctx context.Context, alerts []*types.Aler
 		}
 		sn.log.Warn("Truncated title", "key", key, "max_runes", slackMaxTitleLenRunes)
 	}
+	if tmplErr != nil {
+		sn.log.Warn("failed to template Slack title", "error", tmplErr.Error())
+		tmplErr = nil
```
titolins, line 320: The only other place (after this) I can see where `tmplErr` is referenced is [here](https://github.com/grafana/alerting/pull/248/files#diff-6a83ff1168035e0a58c838a0b2a87e9cd9950ad60e16d801c2d20e45cfff1714R356-R358), which just does a warn log also. I think we could remove that and skip setting it to `nil` here. WDYT?

santihernandezc: That wouldn't work. If we have an error here and skip setting it to nil, the fields we try to template after this one will be blank, which is causing the `invalid_arguments` errors.
The important thing about this PR is that we're clearing the error before templating the `channel` field.

titolins: Ahhh, I didn't actually paid attention to the `TmplText` implementation - they save the err reference there.. wth..

Ok, makes sense 👍

titolins (approved):

PR #249, opened 2024-09-26
Add helper method to decrypt secure settings

santihernandezc:
This PR adds a helper method to `PostableGrafanaReceiver` that returns a map of decoded and decrypted secure settings.
This can be used before sending requests to the Mimir Alertmanager (testing receivers, updating configuration).

### DIFF
--- a/definition/alertmanager.go
+++ b/definition/alertmanager.go
@@ -1,6 +1,7 @@
 package definition
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"reflect"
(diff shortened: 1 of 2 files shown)


definition/alertmanager.go
```diff
@@ -606,3 +607,22 @@ func (r *PostableApiReceiver) GetName() string {
 type PostableGrafanaReceivers struct {
 	GrafanaManagedReceivers []*PostableGrafanaReceiver `yaml:"grafana_managed_receiver_configs,omitempty" json:"grafana_managed_receiver_configs,omitempty"`
 }
+
+// DecryptSecureSettings returns a map containing the decoded and decrypted secure settings.
+func (pgr *PostableGrafanaReceiver) DecryptSecureSettings(decryptFn func(payload []byte) ([]byte, error)) (map[string]string, error) {
```
titolins, line 612: Would be nice to have a unit test for this, just for completeness

titolins (approved):

PR #256, opened 2024-11-27
fix: update slack image upload to use new API

moustafab:
The old `files.upload` API is deprecated and will be removed soon. It doesn't work for new slack workspaces created after May 2024.

The new flow requires a 3 step process:
1. Call `files.getUploadURLExternal` to get the upload URL and file ID.
2. Upload the file to the URL using a PUT request.
3. Call `files.completeUploadExternal` to complete the upload and link it to the thread.

Addresses #255, #235, and https://github.com/grafana/grafana/issues/93398


rwwiv (approved):
LGTM! :shipit:

PR #257, opened 2024-11-29
SNS receiver: truncate subject in case it's over 100 characters

titolins:
## Context
- As per https://github.com/grafana/support-escalations/issues/13632, the SNS API limits the `Subject` field to 100 characters
- Given we allow templating for this, it can be tricky for customers to make sure they're not going over it
- We also truncate the message according to the SNS API limits, so it makes sense to follow this behavior for the `Subject`

### DIFF
--- a/receivers/sns/sns.go
+++ b/receivers/sns/sns.go
@@ -160,7 +160,7 @@ func (s *Notifier) createPublishInput(ctx context.Context, tmpl func(string) str
 
 	messageToSend, isTrunc, err := validateAndTruncateString(tmpl(s.settings.Message), messageSizeLimit)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("message validation failed: %v", err)
 	}
 	if isTrunc {
 		// If we truncated the message we need to add a message attribute showing that it was truncated.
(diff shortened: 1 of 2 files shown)


receivers/sns/sns.go
```diff
@@ -158,34 +158,41 @@ func (s *Notifier) createPublishInput(ctx context.Context, tmpl func(string) str
...
 	if isTrunc {
 		// If we truncated the message we need to add a message attribute showing that it was truncated.
 		messageAttributes["truncated"] = &sns.MessageAttributeValue{DataType: aws.String("String"), StringValue: aws.String("true")}
 	}
 
-	publishInput.SetMessage(messageToSend)
-	publishInput.SetMessageAttributes(messageAttributes)
-
-	subject := tmpl(s.settings.Subject)
+	subject, subjIsTrunc, err := validateAndTruncateString(tmpl(s.settings.Subject), 100)
```
moustafab: Nit: it could be useful to know what the magic number is, perhaps a constant would make sense and can be used in the tests as well?

moustafab (approved):
Looks good to me, small nit

Aga92-92 (approved):

PR #258, opened 2024-12-02
Alertmanager: add sent by to default email template

titolins:
- The remote AM is not adding the build version to the email settings as it should.
- https://github.com/grafana/mimir/pull/9987 was doing that, but the result is misleading
- For that reason, we've decided to instead add a new parameter do the footer text can be customized

Part of: https://github.com/grafana/alerting-squad/issues/873

### DIFF
--- a/receivers/templates/ng_alert_notification.html
+++ b/receivers/templates/ng_alert_notification.html
@@ -1260,11 +1260,7 @@
                   <tbody>
                     <tr>
                       <td align="center" class="txt" style="font-size:0px;padding:10px 25px;word-break:break-word;">
-                        {{ if .SentBy }}
                         <div style="font-family: Inter, Helvetica, Arial; font-size: 13px; line-height: 150%; text-align: center; color: #000000;">&copy; {{ now | date "2006" }} Grafana Labs. Sent by <a href="{{ .AppUrl }}" style="color: #6E9FFF;">{{ .SentBy }}</a>.</div>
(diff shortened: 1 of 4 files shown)


receivers/templates/ng_alert_notification.html
```diff
@@ -1260,7 +1260,11 @@ <h3>✅ {{ .Alerts.Resolved | len }} resolved instances</h3>
                   <tbody>
                     <tr>
                       <td align="center" class="txt" style="font-size:0px;padding:10px 25px;word-break:break-word;">
+                        {{ if .SentBy }}
+                        <div style="font-family: Inter, Helvetica, Arial; font-size: 13px; line-height: 150%; text-align: center; color: #000000;">&copy; {{ now | date "2006" }} Grafana Labs. Sent by <a href="{{ .AppUrl }}" style="color: #6E9FFF;">{{ .SentBy }}</a>.</div>
+                        {{ else }}
```
santihernandezc: Perhaps an easier option would be to remove the `Version` field and use `SentBy` in Grafana and Mimir. I'm fine with either approach really.

santihernandezc (approved):

receivers/templates/ng_alert_notification.html
titolins: Yeah, this is just an overly conservative approach. I'll remove it so we don't need to clean it up later 👍

PR #260, opened 2025-01-10
Add function that returns all default builtin template definitions

JacobsonMT:
Adds a function that computes and returns the default builtin templates that are available for every notifier. This is not used anywhere in this repo but will be used by `grafana/grafana` to expose the defaults in the UI/API.

There is sadly not a simple way to export a parsed template so I opted to capture the combined template during parsing (a similar approach is used in template testing). The main reason for the complexity is that there can be (and are) overlapping template names within the various files that combine into our default definitions, so other options have issues such as:

- **Simply append the text together**: This will result in an invalid template since duplicate names are not allowed within a single "file".
- **Return multiple default template files**: This would be simple in `grafana/alerting` but either requires `grafana/grafana` to return multiple defaults in the API response or requires them to do the work to combine/deduplicate the templates. Neither of these outcomes are particularly good.

**Note**: There is currently a bug in GO template `.String()` for `PipeNode` that assumes all variables are declared. This is fixed in v1.24 with https://github.com/golang/go/commit/6fea4094242fe4e7be8bd7ec0b55df9f6df3f025. The effect of this is that `__text_values_list` will be incorrectly rendered back into template text, so a temporary fix is hard-coded for this specific definition.

### DIFF
--- a/templates/default_template_test.go
+++ b/templates/default_template_test.go
@@ -3,12 +3,14 @@ package templates
 import (
 	"context"
 	"net/url"
+	"strings"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
+
 	"github.com/prometheus/alertmanager/types"
 	"github.com/prometheus/common/model"
-	"github.com/stretchr/testify/require"
 
 	"github.com/grafana/alerting/logging"
 )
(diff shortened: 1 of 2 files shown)


yuri-tceretian (approved):
LGTM!

PR #264, opened 2025-01-13
Add NoopDecode function for non-base64-encoded secrets

santihernandezc:
When the Mimir Alertmanager uses Grafana integrations, secrets are not base64-encoded. We rely on an [error returned from the decoding function](https://github.com/grafana/alerting/blob/267368fd1968774d47b9553d113f9438eeeeed0c/notify/receivers.go#L247-L254) to tell whether secrets are base64-decoded or not (check [here](https://github.com/grafana/alerting/blob/267368fd1968774d47b9553d113f9438eeeeed0c/notify/receivers.go#L249-L253)). This works fine in most cases, but some non-base64-encoded strings will still be valid base64. In these cases, the decoding function will apply base64-decoding rules and output gibberish, thus breaking integrations.

This PR adds a `NoopDecode` function to be used by the Mimir Alertmanager. This function converts a `map[string]string` into a `map[string][]byte` without trying to decode the values.

Related Mimir PR: https://github.com/grafana/mimir/pull/10427

### DIFF
--- a/notify/receivers_test.go
+++ b/notify/receivers_test.go
@@ -86,7 +86,7 @@ func TestBuildReceiverConfiguration(t *testing.T) {
 			counter++
 			return decrypt(ctx, sjd, key, fallback)
 		}
-		_, _ = BuildReceiverConfiguration(context.Background(), recCfg, decryptCount)
+		_, _ = BuildReceiverConfiguration(context.Background(), recCfg, DecodeSecretsFromBase64, decryptCount)
 		require.Greater(t, counter, 0)
 	})
 	t.Run("should fail if at least one config is invalid", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


notify/receivers_test.go
```diff
@@ -113,13 +113,13 @@ func TestBuildReceiverConfiguration(t *testing.T) {
...
 		recCfg := &APIReceiver{ConfigReceiver: ConfigReceiver{Name: "test-receiver"}}
-		parsed, err := BuildReceiverConfiguration(context.Background(), recCfg, decrypt)
+		parsed, err := BuildReceiverConfiguration(context.Background(), recCfg, DecodeSecretsFromBase64, decrypt)
 		require.NoError(t, err)
 		require.Equal(t, recCfg.Name, parsed.Name)
 	})
 	t.Run("should support non-base64-encoded secrets", func(t *testing.T) {
 		recCfg := &APIReceiver{ConfigReceiver: ConfigReceiver{Name: "test-receiver"}}
-		invalidBase64 := "not encoded"
+		invalidBase64 := "test"
```
santihernandezc, line 122: The string "test" is valid base64, the test will fail if the `DecodeSecretsFromBase64` decoding function is used.

yuri-tceretian (approved):
LGTM as a quick fix to stop the bleeding.

PR #265, opened 2025-01-14
Log duplicated receivers

yuri-tceretian:
The Grafana Alertmanager declares receivers as a list of objects, each with a name attribute. When Alertmanager is configured, the receivers are grouped into a map where the key is the name of the receiver. 
Therefore, if Alertmanager configuration contains many receivers under the same name, which was possible when the configuration was submitted via now-legacy raw API, the last receiver in the list - wins. 
See https://github.com/grafana/grafana/issues/67330

Since introducing a new API for receivers (see feature flag alertingApiServer) this situation should not be possible anymore. 

This PR adds logging to simplify the identification of bad configurations.

### DIFF
--- a/notify/grafana_alertmanager.go
+++ b/notify/grafana_alertmanager.go
@@ -18,8 +18,6 @@ import (
 	"github.com/go-openapi/strfmt"
 	"golang.org/x/sync/errgroup"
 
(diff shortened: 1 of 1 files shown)


JacobsonMT (approved):
LGTM

PR #266, opened 2025-01-16
Alerting: Use rule uid in silenceURL template instead of alertname and folder

JacobsonMT:
When possible use rule uid instead of alertname and grafana_folder when templating the silenceURL as this has better compatibility with minimal permission RBAC users.

Example before: http://localhost:3000/alerting/silence/new?alertmanager=grafana&matcher=alertname%3Dalways+firing+screenshot&matcher=grafana_folder%3Dtest&orgId=1 
Example after: http://localhost:3000/alerting/silence/new?alertmanager=grafana&matcher=__alert_rule_uid__%3Dae9zjkfxiyk1sa&orgId=1

Relates to: https://github.com/grafana/grafana/issues/99115

### DIFF
--- a/receivers/pushover/pushover_test.go
+++ b/receivers/pushover/pushover_test.go
@@ -11,11 +11,12 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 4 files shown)


yuri-tceretian (approved):
LGTM! Great change!

PR #267, opened 2025-01-17
Alerting: Refactor extendAlert function

JacobsonMT:
Clean it up so that it requires less mutation of a single `url.URL` struct to create the various extended field URLs.

- Now uses a `baseURL` + shallow copying in separate generator functions. (b1857e9eea0d7e1117be7c60ce63d546bb024cc5)
- Also reorders the field calculations so that parse errors don't prevent other fields from being populated. Ex. Failing to parse the `baseURL` shouldn't prevent us from adding `Annotations` or the adding the `orgID` to `generatorURL`. (a6271361fe881b9223bc5f2d8c9d0b06addcd2de)

### DIFF
--- a/templates/template_data.go
+++ b/templates/template_data.go
@@ -173,7 +173,7 @@ func extendAlert(alert template.Alert, externalURL string, logger log.Logger) *E
 	}
 
 	if generatorURL, err := url.Parse(extended.GeneratorURL); err != nil {
-		level.Debug(logger).Log("msg", "failed to parse generator URL while extending template data", "url", extended.GeneratorURL, "error", err.Error())
+		level.Warn(logger).Log("msg", "failed to parse generator URL while extending template data", "url", extended.GeneratorURL, "error", err.Error())
 	} else if orgID := alert.Annotations[models.OrgIDAnnotation]; len(orgID) > 0 {
 		// Refactor note: We only modify the URL if there is something to add. Otherwise, the original string is kept.
 		setQueryParam(generatorURL, "orgId", orgID)
@@ -187,7 +187,7 @@ func extendAlert(alert template.Alert, externalURL string, logger log.Logger) *E
 			}
 		}
 
-		// TODO: Remove in Grafana 10
+		// TODO: Remove in Grafana 12
 		extended.ValueString = alert.Annotations[models.ValueStringAnnotation]
 	}
 
@@ -197,7 +197,7 @@ func extendAlert(alert template.Alert, externalURL string, logger log.Logger) *E
 	}
 	baseURL, err := url.Parse(externalURL)
 	if err != nil {
-		level.Debug(logger).Log("msg", "failed to parse external URL while extending template data", "url", externalURL, "error", err.Error())
+		level.Warn(logger).Log("msg", "failed to parse external URL while extending template data", "url", externalURL, "error", err.Error())
 		return extended
 	}
 
--- a/templates/template_data.go
+++ b/templates/template_data.go
@@ -172,30 +172,12 @@ func extendAlert(alert template.Alert, externalURL string, logger log.Logger) *E
 		Fingerprint:  alert.Fingerprint,
 	}
 
-	// fill in some grafana-specific urls
-	if len(externalURL) == 0 {
-		return extended
-	}
-	baseURL, err := url.Parse(externalURL)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


templates/template_data.go
```diff
@@ -172,69 +172,80 @@ func extendAlert(alert template.Alert, externalURL string, logger log.Logger) *E
 		Fingerprint:  alert.Fingerprint,
 	}
 
+	if generatorURL, err := url.Parse(extended.GeneratorURL); err != nil {
+		level.Debug(logger).Log("msg", "failed to parse generator URL while extending template data", "url", extended.GeneratorURL, "error", err.Error())
```
yuri-tceretian: Let's make it warn? I think this is important enough problem

templates/template_data.go
```diff
@@ -172,69 +172,80 @@ func extendAlert(alert template.Alert, externalURL string, logger log.Logger) *E
...
+
+	if alert.Annotations != nil {
+		if s, ok := alert.Annotations[models.ValuesAnnotation]; ok {
+			if err := json.Unmarshal([]byte(s), &extended.Values); err != nil {
+				level.Warn(logger).Log("msg", "failed to unmarshal values annotation", "error", err.Error())
+			}
+		}
+
+		// TODO: Remove in Grafana 10
+		extended.ValueString = alert.Annotations[models.ValueStringAnnotation]
```
yuri-tceretian, line 191: Lol, it's Grafana 12 soon and we still have it! Perhaps change todo?

templates/template_data.go
```diff
@@ -172,69 +172,80 @@ func extendAlert(alert template.Alert, externalURL string, logger log.Logger) *E
...
+	}
+
 	// fill in some grafana-specific urls
 	if len(externalURL) == 0 {
 		return extended
 	}
-	u, err := url.Parse(externalURL)
+	baseURL, err := url.Parse(externalURL)
 	if err != nil {
 		level.Debug(logger).Log("msg", "failed to parse external URL while extending template data", "url", externalURL, "error", err.Error())
```
yuri-tceretian: Same about warning here

yuri-tceretian (approved):
LGTM, one thing still bothers me is function `setQueryParam` that basically parses the current query string, mutates it and serializes back to string. It's not a huge deal and I do not think the impact on performance is substantial enough to consider refactoring, but every time I see I my hands itching.
WDYT about rewriting it to some URLBuilder struct like? 
```go
 // URLBuilder struct encapsulates URL logic and maintains query parameters as a map
type URLBuilder struct {
	BaseURL    *url.URL
	QueryParams map[string]string
}

// NewURLBuilder creates a new instance of URLBuilder
func NewURLBuilder(base string) (*URLBuilder, error) {
	parsedURL, err := url.Parse(base)
	if err != nil {
		return nil, err
	}
	return &URLBuilder{
		BaseURL:    parsedURL,
		QueryParams: make(map[string]string),
	}, nil
}

// SetQueryParam sets a query parameter
func (b *URLBuilder) SetQueryParam(key, value string) {
	b.QueryParams[key] = value
}

// Build constructs the final URL with all query parameters
func (b *URLBuilder) Build() string {
	query := url.Values{}
	for key, value := range b.QueryParams {
		query.Set(key, value)
	}
	b.BaseURL.RawQuery = query.Encode()
	return b.BaseURL.String()
}
```

templates/template_data.go
JacobsonMT: Not sure this one is important enough to Warn on but I'll change it as it seems correct in context. The other one for `baseURL` should definitely be Warn though.

JacobsonMT:
> LGTM, one thing still bothers me is function `setQueryParam` that basically parses the current query string, mutates it and serializes back to string. It's not a huge deal and I do not think the impact on performance is substantial enough to consider refactoring, but every time I see I my hands itching. WDYT about rewriting it to some URLBuilder struct like?


Yah the abstraction on URL feels like it needs something that prevents the odd repeated parse+encode pattern. However, the current implementation encodes the query parameters the minimal number of times from what I can tell:
1. `GeneratorURL` to add `orgID`.
2. `BaseURL`/`DashboardURL` to add `orgID`.
3. `PanelURL` to add `viewPanel`
4. `SilenceURL` to add `matchers`

The `URLBuilder`, while I agree less itchy :), won't improve that aspect at least.

PR #268, opened 2025-01-22
Add experimental stage that will catch race conditions between nflog and pipeline

yuri-tceretian:
This PR adds a new stage to the pipeline that is similar to DedupStage but it checks that the timestamp when pipeline was flushed matches the timestamp of the state: if latter is greater than the former then it is indicator that the pipeline was already executed.

It provides the ability to configure how to react in this case: just log or stop the pipeline. This will be controlled by a feature flag in Grafana.

This is a non-intrusive way to estimate how often the situation described in 
https://github.com/prometheus/alertmanager/pull/3283/files
happens in reality and the ability to stop the notification pipeline in that case.

### DIFF
--- a/notify/grafana_alertmanager.go
+++ b/notify/grafana_alertmanager.go
@@ -36,8 +36,8 @@ import (
 	"github.com/prometheus/common/model"
 
 	"github.com/grafana/alerting/cluster"
-	"github.com/grafana/alerting/notify/experimental"
 	"github.com/grafana/alerting/notify/nfstatus"
+	"github.com/grafana/alerting/notify/pipeline"
 
 	"github.com/grafana/alerting/models"
 	"github.com/grafana/alerting/templates"
@@ -899,7 +899,7 @@ func (am *GrafanaAlertmanager) createReceiverStage(name string, integrations []*
 		s = append(s, notify.NewWaitStage(wait))
 		s = append(s, notify.NewDedupStage(integrations[i], notificationLog, recv))
 		if act == LogOnly || act == StopPipeline {
-			s = append(s, experimental.NewPipelineAndStateTimestampCoordinationStage(notificationLog, recv, act == StopPipeline))
+			s = append(s, pipeline.NewPipelineAndStateTimestampCoordinationStage(notificationLog, recv, act == StopPipeline))
 		}
 		s = append(s, notify.NewRetryStage(integrations[i], name, am.stageMetrics))
 		s = append(s, notify.NewSetNotifiesStage(notificationLog, recv))
--- a/notify/grafana_alertmanager.go
+++ b/notify/grafana_alertmanager.go
@@ -158,7 +158,16 @@ type Configuration interface {
 
 	Hash() [16]byte
 	Raw() []byte
+	PipelineAndSateTimestampsMismatchAction() PipelineAndSateTimestampsMismatchAction
 }
+
+type PipelineAndSateTimestampsMismatchAction string
+
+var (
+	Disabled     PipelineAndSateTimestampsMismatchAction = "disabled"
+	LogOnly      PipelineAndSateTimestampsMismatchAction = "log-only"
+	StopPipeline PipelineAndSateTimestampsMismatchAction = "stop-pipeline"
+)
 
 type Limits struct {
 	MaxSilences         int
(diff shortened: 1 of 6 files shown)


notify/grafana_alertmanager.go
```diff
@@ -157,8 +158,17 @@ type Configuration interface {
 
 	Hash() [16]byte
 	Raw() []byte
+	PipelineAndSateTimestampsMismatchAction() PipelineAndSateTimestampsMismatchAction
 }
 
+type PipelineAndSateTimestampsMismatchAction string
```
titolins: typo
```suggestion
type PipelineAndStateTimestampsMismatchAction string
```

notify/grafana_alertmanager.go
```diff
@@ -157,8 +158,17 @@ type Configuration interface {
...
+
+var (
+	Disabled     PipelineAndSateTimestampsMismatchAction = "disabled"
+	LogOnly      PipelineAndSateTimestampsMismatchAction = "log-only"
+	StopPipeline PipelineAndSateTimestampsMismatchAction = "stop-pipeline"
```
titolins: Considering this is in `notify` package, it might make sense to be more specific in the variable names. E.g. Following the same pattern as the type (or shorter might be better). WDYT?
```suggestion
	PipelineAndStateTimestampsMismatchDisabled     PipelineAndStateTimestampsMismatchAction = "disabled"
	PipelineAndStateTimestampsMismatchLogOnly      PipelineAndStateTimestampsMismatchAction = "log-only"
	PipelineAndStateTimestampsMismatchLogStop PipelineAndStateTimestampsMismatchAction = "stop-pipeline"
```

notify/grafana_alertmanager.go
```diff
@@ -888,6 +898,9 @@ func (am *GrafanaAlertmanager) createReceiverStage(name string, integrations []*
 		var s notify.MultiStage
 		s = append(s, notify.NewWaitStage(wait))
 		s = append(s, notify.NewDedupStage(integrations[i], notificationLog, recv))
+		if act == LogOnly || act == StopPipeline {
```
titolins: nit:
```suggestion
		if act != Disabled {
```

notify/grafana_alertmanager.go
```diff
@@ -157,8 +158,17 @@ type Configuration interface {
 
 	Hash() [16]byte
 	Raw() []byte
+	PipelineAndSateTimestampsMismatchAction() PipelineAndSateTimestampsMismatchAction
```
titolins: Just curious if you've considered adding this to the [GrafanaAlertmanager](https://github.com/grafana/alerting/blob/main/notify/grafana_alertmanager.go#L69-L69) struct instead. Then the flag could be injected during [initialization](https://github.com/grafana/grafana/blob/main/pkg/services/ngalert/notifier/alertmanager.go#L138-L138)

notify/pipeline/coordination_stage.go
```diff
@@ -0,0 +1,64 @@
+package pipeline
```
titolins: Would't it make sense to move this to [prometheus-alertmanager](https://github.com/grafana/prometheus-alertmanager/blob/main/notify/notify.go#L1-L1)?

yuri-tceretian: Later, when it becomes less experimental :)

notify/grafana_alertmanager.go
yuri-tceretian: in Go there is no strongly typed enums. So, value of the parameter could be any string. I perfer to react on known values.

notify/grafana_alertmanager.go
yuri-tceretian: Agree

notify/pipeline/coordination_stage.go
yuri-tceretian: Currently, I do not want to add anything to the fork because it may cause problems syncing it with upstream

notify/grafana_alertmanager.go
yuri-tceretian: It looked simpler but no other reason. Will move to AM

notify/pipeline/coordination_stage.go
titolins: Ah, I see. Makes sense 👍

notify/grafana_alertmanager.go
```diff
@@ -191,16 +197,17 @@ func (c *GrafanaAlertmanagerConfig) Validate() error {
...
+		logger:                                  log.With(logger, "component", "alertmanager", tenantKey, tenantID),
+		marker:                                  types.NewMarker(m.Registerer),
+		stageMetrics:                            notify.NewMetrics(m.Registerer, featurecontrol.NoopFlags{}),
+		dispatcherMetrics:                       dispatch.NewDispatcherMetrics(false, m.Registerer),
+		peer:                                    peer,
+		peerTimeout:                             config.PeerTimeout,
+		Metrics:                                 m,
+		tenantID:                                tenantID,
+		externalURL:                             config.ExternalURL,
+		pipelineAndSateTimestampsMismatchAction: config.PipelineAndStateTimestampsMismatchAction,
```
titolins: typo got copied here 😅 

```suggestion
		pipelineAndStateTimestampsMismatchAction: config.PipelineAndStateTimestampsMismatchAction,
```

titolins (approved):
🚀

notify/grafana_alertmanager.go
yuri-tceretian: I can bet I fixed it!!

PR #271, opened 2025-01-29
Fix UTF-8 not allowed in Equal field for inhibition rules

grobinson-grafana:
This pull request fixes a bug where UTF-8 characters are not allowed in the Equal field for inhibition rules, even when UTF-8 strict mode is enabled.

This bug occurred because we forgot to override the validation in model.LabelName. I have copied the same logic used for GroupBy with GroupByStr, adding EqualStr.

It also adds tests for decoding inhibition rules as JSON and YAML, to make sure that the new EqualsStr field is decoded into Equals. See prometheus/alertmanager#4177.


grobinson-grafana:
This will pass when https://github.com/grafana/prometheus-alertmanager/pull/94 is merged and `go.mod` is updated.

titolins (approved):

santihernandezc (approved):

PR #273, opened 2025-01-31
Add base setup for integration tests

titolins:
## Description
- In the context of the flappy alerts investigation, we're trying to create a minimal setup that would help us replicate the issue and therefore validate possible solutions
- This is just a base as of now, providing the tools to setup a Grafana cluster and provision the required resources
- The idea as of now is to capture a single case - i.e. once the alert is resolved, there should be no more firing notifications
- I'll run this against https://github.com/grafana/grafana/pull/99825 to see if we're able to mitigate at least some of the issues raised in escalations

## Setup
- It can be run with make - e.g.: `GRAFANA_VERSION=steady make integration-tests`
- This will build and run the following containers:
  - postgres;
  - loki (used for the alert state history);
  - webhook-receiver (simple webhook application that records alert notifications and exposes a GET endpoint to retrieve them); and
  - Grafana (cluster size is passed to the `AlertmanagerScenario.Start` function)
- It will also provision an alert rule (plus all other required resources) using a predictable csv pulse (from testdata datasource)
- The predictable series will cause the alert to trigger and resolve multiple times
- The test times out after 10 minutes and is set to fail so we see the log output (we query the webhook receiver every 15s)
- Having a long timeout allows us to also access the grafana instances and query the webhook `/notifications` API directly

## Example
- Attached below, there are screenshots of the query and the alert history page and the response from the `/notifications` API after around 5 minutes of tests running
- We can see that even though the alert triggered once, we got 3 firing notifications and 2 resolved before the test was halted
- The first 2 notifications (1 firing then 1 resolved) are 1 minute apart, but then the other 3 are each 15s apart (which is exactly the value configured for the peer timeout)
Screenshot 2025-01-31 at 00 43 57
Screenshot 2025-01-31 at 00 44 02

```json
{
	"history": [
		{
			"status": "firing",
			"timeNow": "2025-01-30T23:42:30.02673093Z",
			"startsAt": "2025-01-30T23:42:00Z",
			"node": "172.20.0.5",
			"deltaLastSeconds": 0,
			"deltaStartSeconds": 184.557485293
		},
		{
			"status": "resolved",
			"timeNow": "2025-01-30T23:43:30.025678722Z",
			"startsAt": "2025-01-30T23:42:00Z",
			"node": "172.20.0.5",
			"deltaLastSeconds": 59.999020776,
			"deltaStartSeconds": 244.556506069
		},
		{
			"status": "firing",
			"timeNow": "2025-01-30T23:43:45.030064131Z",
			"startsAt": "2025-01-30T23:42:00Z",
			"node": "172.20.0.10",
			"deltaLastSeconds": 15.004286007,
			"deltaStartSeconds": 259.560792076
		},
		{
			"status": "resolved",
			"timeNow": "2025-01-30T23:44:00.030662888Z",
			"startsAt": "2025-01-30T23:42:00Z",
			"node": "172.20.0.7",
			"deltaLastSeconds": 15.000598715,
			"deltaStartSeconds": 274.561390791
		},
		{
			"status": "firing",
			"timeNow": "2025-01-30T23:44:15.031228673Z",
			"startsAt": "2025-01-30T23:42:00Z",
			"node": "172.20.0.12",
			"deltaLastSeconds": 15.000283257,
			"deltaStartSeconds": 289.561674048
		}
	],
	"stats": {
		"firing": 3,
		"resolved": 2
	}
}
```


integration/grafana.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/grafana/e2e"
+	gapi "github.com/grafana/grafana-api-golang-client"
+	"gopkg.in/yaml.v3"
+)
+
+//go:embed grafana-versions.yaml
+var grafanaVersionsJson []byte
+
+const (
+	defaultGrafanaImage = "grafana/grafana-enterprise-dev:10.1.0-58862pre"
```
yuri-tceretian, line 18: I think it makes sense to just require an image set via envvars unless there is a good reason to refer to exactly that version

yuri-tceretian: Ah, it's not used anywhere. Let's remove it.

titolins:
Closed in favor of https://github.com/grafana/grafana/pull/100106

integration/grafana.go
titolins: Removed in https://github.com/grafana/grafana/pull/100106

PR #275, opened 2025-02-01
Add EmbeddedContents and use in email notifier

JacobsonMT:
Adds support for embedding `[]byte` in `defaultEmailSender` instead of filenames. This is backwards compatible as it uses a new field `EmbeddedContents` to add an alternative to the existing `EmbeddedFiles` which takes filenames.

Also, uses this new field in the email notifier. 

This is needed because of upcoming changes to the image `Provider` abstraction. It will only expose `[]byte` not files or filenames. In other words, integrations no longer know that the data is coming from disk or not, so filenames and paths are leaked implementation details.

This should be noop functionally.

This will requires corresponding implementation in `grafana/grafana` for local alertmanager: https://github.com/grafana/grafana/pull/99937

<details>
  <summary>Embed confirmed still working</summary>

  image

</details>

### DIFF
--- a/receivers/email/email.go
+++ b/receivers/email/email.go
@@ -58,23 +58,19 @@ func (en *Notifier) Notify(ctx context.Context, alerts ...*types.Alert) (bool, e
 	}
 
 	// Extend alerts data with images, if available.
-	embeddedFiles := make(map[string]io.Reader)
-	defer func() {
-		for _, reader := range embeddedFiles {
-			if closer, ok := reader.(io.Closer); ok {
-				closer.Close()
-			}
-		}
-	}()
+	embeddedContents := make([]receivers.EmbeddedContent, 0)
 	_ = images.WithStoredImages(ctx, en.log, en.images,
 		func(index int, image images.Image) error {
 			if len(image.URL) != 0 {
 				data.Alerts[index].ImageURL = image.URL
 			} else if len(image.Path) != 0 {
-				if f, err := openFile(image.Path); err == nil {
+				if b, err := readFile(image.Path); err == nil {
 					name := filepath.Base(image.Path)
 					data.Alerts[index].EmbeddedImage = name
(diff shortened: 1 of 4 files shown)


receivers/email/email.go
```diff
@@ -108,3 +112,21 @@ func (en *Notifier) Notify(ctx context.Context, alerts ...*types.Alert) (bool, e
 func (en *Notifier) SendResolved() bool {
 	return !en.GetDisableResolveMessage()
 }
+
+func readFile(path string) ([]byte, error) {
+	_, err := os.Stat(path)
```
yuri-tceretian: I think this check here is redundant because Open below will catch it too. Perhaps remove it?

receivers/email/email.go
```diff
@@ -108,3 +112,21 @@ func (en *Notifier) Notify(ctx context.Context, alerts ...*types.Alert) (bool, e
...
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+
+	defer func() {
+		_ = f.Close()
+	}()
+
+	return io.ReadAll(f)
```
yuri-tceretian, line 126: WDYT about checking file size and creating a hard limit for the size of file? We can take double of what render service produces.

yuri-tceretian (commented):
LGTM

receivers/email/email.go
JacobsonMT: Ah, true. Didn't put too much thought into this function when moving it out of `Notify` as it's immediately going away in the next PR. I'll remove the `os.Stat`

receivers/email/email.go
JacobsonMT: This function is going away in the next PR, it's just a stop-gap. So, we can discuss adding something like this in the next one IMO.

yuri-tceretian (approved):

PR #276, opened 2025-02-03
New image Provider abstraction to allow for remote AM screenshots

JacobsonMT:
### Overview

This PR reworks the image `Provider` interface so that we can get screenshots in notifications working in both local and remote AMs. To this end, it:

- Simplifies the image `Provider` abstraction to hide Path and File information.
- Modifies a few integrations that use image upload to use the new `Image` struct.
- Provides two generic Provider implementations:
  - `TokenProvider`: Provider that takes tokens from `__alertImageToken__` annotation and attempts to retrieve an Image from the attached `TokenStore`. Allows fetching raw data. This is used by local AM where raw data is fetches from disk, but could eventually be used in remote AM with an appropriate image store.
  - `URLProvider`: Provider that takes urls from `__alert_image_url__` annotation and uses this as the public image url. Fetching raw data is unsupported, thus integrations that upload are not supported for screenshots. This is meant to be used by remote AM as a store is not necessary.
 
New abstraction is:

```go
type Provider interface {
	// GetImage takes an alert and returns its associated image.
	GetImage(ctx context.Context, alert types.Alert) (*Image, error)
}

type Image struct {
	// URL is the public URL of the image. This URL should not be treated as a trusted source and should not be
	// downloaded directly. RawData should be used to retrieve the image data.
	URL string
	// RawData returns the raw image data. Depending on the provider, this may be a file read, a network request, or
	// unsupported. It's the responsibility of the Provider to ensure that the data is safe to read.
	RawData func(ctx context.Context) (ImageContent, error)
}

type ImageContent struct {
	// Name is the unique identifier for the image. Usually this will be an image filename, but is not required to be.
	Name string
	// Content is the raw image data.
	Content []byte
}
```

### Other Notes

- Raw data fetching is encapsulated in the `Image` struct as `RawData(...)` so that we don't need to unnecessarily fetch from the `TokenStore` twice for those integrations that attempt both public URL and image upload/embed.
- Lots of defined errors are no longer needed with this simpler abstraction. Notably, we return `nil` when image isn't found instead of an error for conditional flow.

Split by easy to understand commit for those that prefer reviewing by commit.

### Image upload testing:

<details>
  <summary>Slack</summary>

Screenshot from 2025-02-19 10-33-47

</details>

<details>
  <summary>Email</summary>

Screenshot from 2025-02-19 10-30-24

</details>

<details>
  <summary>Discord</summary>

Screenshot from 2025-02-19 10-29-16

</details>

<details>
  <summary>Telegram</summary>

Screenshot from 2025-02-19 10-28-32

</details>

### DIFF
--- a/images/images.go
+++ b/images/images.go
@@ -15,16 +15,7 @@ var (
 	// the maximum number of images has been iterated.
 	ErrImagesDone = errors.New("images done")
 
-	// ErrImagesNoPath is returned whenever an image is found but has no path on disk.
-	ErrImagesNoPath = errors.New("no path for image")
-
-	// ErrImagesNoURL is returned whenever an image is found but has no URL.
-	ErrImagesNoURL = errors.New("no URL for image")
-
 	ErrImagesUnavailable = errors.New("alert screenshots are unavailable")
-
-	// ErrNoImageForAlert is returned when no image is associated to a given alert.
-	ErrNoImageForAlert = errors.New("no image for alert")
 )
 
 type ImageContent struct {
--- a/images/images.go
+++ b/images/images.go
@@ -3,8 +3,6 @@ package images
 import (
 	"context"
 	"errors"
-	"io"
-	"time"
 
 	"github.com/prometheus/alertmanager/types"
 )
@@ -29,11 +27,20 @@ var (
 	ErrNoImageForAlert = errors.New("no image for alert")
 )
 
+type ImageContent struct {
+	// Name is the unique identifier for the image. Usually this will be an image filename, but is not required to be.
+	Name string
+	// Content is the raw image data.
+	Content []byte
+}
+
 type Image struct {
-	Token     string
-	Path      string
-	URL       string
-	CreatedAt time.Time
+	// URL is the public URL of the image. This URL should not be treated as a trusted source and should not be
+	// downloaded directly. RawData should be used to retrieve the image data.
+	URL string
+	// RawData returns the raw image data. Depending on the provider, this may be a file read, a network request, or
+	// unsupported. It's the responsibility of the Provider to ensure that the data is safe to read.
+	RawData func(ctx context.Context) (ImageContent, error)
 }
 
 func (i Image) HasURL() bool {
@@ -41,24 +48,8 @@ func (i Image) HasURL() bool {
 }
 
 type Provider interface {
-	// GetImage takes a token (identifier) and returns the image that token belongs to.
-	// Returns `ErrImageNotFound` if there's no image for said token.
-	//
-	// Deprecated: This method will be removed when all integrations use GetImageURL and/or GetRawImage,
-	// which allow integrations to get just the data they need for adding images to notifications.
-	// Use any of those two methods instead.
-	GetImage(ctx context.Context, token string) (*Image, error)
-
-	// GetImageURL returns the URL of an image associated with a given alert.
-	//	- Returns `ErrImageNotFound` if no image is found.
-	//	- Returns `ErrImagesNoURL` if the image doesn't have a URL.
-	GetImageURL(ctx context.Context, alert *types.Alert) (string, error)
-
-	// GetRawImage returns an io.Reader to read the bytes of an image associated with a given alert
-	// and a string representing the filename.
-	//	- Returns `ErrImageNotFound` if no image is found.
-	//	- Returns `ErrImagesNoPath` if the image doesn't have a path on disk.
-	GetRawImage(ctx context.Context, alert *types.Alert) (io.ReadCloser, string, error)
+	// GetImage takes an alert and returns its associated image.
+	GetImage(ctx context.Context, alert types.Alert) (*Image, error)
 }
 
 type UnavailableProvider struct{}
--- a/images/images.go
+++ b/images/images.go
@@ -51,18 +51,3 @@ type Provider interface {
 	// GetImage takes an alert and returns its associated image.
 	GetImage(ctx context.Context, alert types.Alert) (*Image, error)
 }
-
-type UnavailableProvider struct{}
-
-// GetImage returns the image with the corresponding token, or ErrImageNotFound.
-func (u *UnavailableProvider) GetImage(context.Context, string) (*Image, error) {
-	return nil, ErrImagesUnavailable
-}
-
-func (u *UnavailableProvider) GetImageURL(context.Context, *types.Alert) (string, error) {
-	return "", ErrImagesUnavailable
-}
-
-func (u *UnavailableProvider) GetRawImage(context.Context, *types.Alert) (io.ReadCloser, string, error) {
-	return nil, "", ErrImagesUnavailable
-}
--- a/images/providers.go
+++ b/images/providers.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 
+	"github.com/grafana/alerting/logging"
 	"github.com/grafana/alerting/models"
 	"github.com/prometheus/alertmanager/types"
 )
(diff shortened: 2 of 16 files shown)


images/providers.go
```diff
@@ -0,0 +1,100 @@
...
+	if url == "" {
+		return nil, nil
+	}
+
+	return &Image{
+		URL: url,
+		RawData: func(_ context.Context) (ImageContent, error) {
+			// Raw images are not available for alerts provided directly by annotations as the image data is non-local.
+			// While it might be possible to download the image data, it's generally not safe to do so as the URL is
+			// not guaranteed to be trusted.
```
santihernandezc: Did you mean...?
```suggestion
			// Raw images are not available for URLs provided directly by annotations as the image data is non-local.
			// While it might be possible to download the image data, it's generally not safe to do so as the URL is
			// not guaranteed to be trusted.
```

images/providers.go
```diff
@@ -0,0 +1,100 @@
...
+	}, nil
+}
+
+type TokenStore interface {
+	GetImage(ctx context.Context, token string) (*Image, error)
+}
+
+// TokenProvider is an implementation of the Image Provider interface that retrieves images from a store
+// using tokens.
+// Image data should be considered trusted by the fact that the stored image url and content and not user-modifiable.
```
santihernandezc: `ImageProvider` without spaces, avoiding the line break before "using tokens", uppercasing "URL", "and" -> "are".
```suggestion
// TokenProvider is an implementation of the ImageProvider interface that retrieves images from a store using tokens.
// Image data should be considered trusted by the fact that the stored image URL and content and not user-modifiable.
```

Or more concisely:
```suggestion
// TokenProvider implements the ImageProvider interface, retrieving images from a store using tokens.
// Image data should be considered trusted as the stored image URL and content are not user-modifiable.
```

images/providers.go
```diff
@@ -0,0 +1,100 @@
...
+	"context"
+	"errors"
+
+	"github.com/grafana/alerting/logging"
+	"github.com/grafana/alerting/models"
+	"github.com/prometheus/alertmanager/types"
+)
+
+// ErrImageUploadNotSupported is returned when image uploading is not supported.
+var ErrImageUploadNotSupported = errors.New("image upload is not supported")
```
santihernandezc, line 14: The comment is a bit redundant. I'd either add more details (e.g. why this can happen) or remove it.

Suggestion: I just know we're going to forget what this error means and where it comes from. We could make it more explicit so if we find it in the logs, we know right away what this is about.
```suggestion
// ErrImageUploadNotSupported is returned when the image is provided as a URL in an annotation,
// as these URLs cannot be trusted for direct download/upload.
var ErrImageUploadNotSupported = errors.New("image upload is not supported when using URLs in annotations")
```

images/providers.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (i TokenProvider) GetImage(ctx context.Context, alert types.Alert) (*Image, error) {
+	token := GetImageToken(alert)
+	if token == "" {
+		return nil, nil
+	}
+
+	// Assume the uri is a token because we used to store tokens as plain strings.
+	logger := i.logger.New("token", token)
+	logger.Debug("Received an image token in annotations")
```
santihernandezc: [`logger.New()`](https://github.com/grafana/grafana/blob/7c8fddc7291f181b83fee6e6a7be4f1190a38229/pkg/infra/log/log.go#L175-L189) creates a new logger struct and returns a pointer. This reference to the logger will escape to the heap and need to be GC'd. Some other allocations are done along the way, as the function also calls some methods from 3rd party libraries.

I would add the context to the original logger and use it.
```suggestion
	i.logger.Debug("Received an image token in annotations", "token", token)
```

Same for the `Info` line below.

images/utils.go
```diff
@@ -45,12 +37,15 @@ func getImage(ctx context.Context, l logging.Logger, imageProvider Provider, ale
 
 // WithStoredImages retrieves the image for each alert and then calls forEachFunc
 // with the index of the alert and the retrieved image struct. If the alert does
-// not have an image token, or the image does not exist then forEachFunc will not be
+// not have an image then forEachFunc will not be
 // called for that alert. If forEachFunc returns an error, WithStoredImages will return
 // the error and not iterate the remaining alerts. A forEachFunc can return ErrImagesDone
 // to stop the iteration of remaining alerts if the intended image or maximum number of
 // images have been found.
```
santihernandezc: Weird line break.
```suggestion
// not have an image then forEachFunc will not be called for that alert.
// If forEachFunc returns an error, WithStoredImages will return the error
// and not iterate the remaining alerts. A forEachFunc can return ErrImagesDone
// to stop the iteration of remaining alerts if the intended image or maximum number of
// images have been found.
```

images/utils.go
```diff
@@ -24,19 +21,14 @@ type forEachImageFunc func(index int, image Image) error
...
 	ctx, cancelFunc := context.WithTimeout(ctx, ProviderTimeout)
 	defer cancelFunc()
 
-	img, err := imageProvider.GetImage(ctx, token)
+	img, err := imageProvider.GetImage(ctx, alert)
 	if errors.Is(err, ErrImageNotFound) || errors.Is(err, ErrImagesUnavailable) {
 		return nil, nil
 	} else if err != nil {
-		l.Warn("failed to get image with token", "token", token, "error", err)
+		l.Warn("failed to get image", "error", err)
```
santihernandezc, line 31: Suggestion for extra context:
```suggestion
		l.Warn("failed to get image", "fingerprint", alert.Fingerprint(), "error", err)
```
Either that or `alert.String()`.

receivers/discord/discord.go
```diff
@@ -219,66 +217,49 @@ func (d Notifier) SendResolved() bool {
...
 func (d Notifier) constructAttachments(ctx context.Context, alerts []*types.Alert, embedQuota int) []discordAttachment {
 	attachments := make([]discordAttachment, 0, embedQuota)
-	embedsUsed := 0
-	for _, alert := range alerts {
-		// Check if the image limit has been reached at the start of each iteration.
-		if embedsUsed >= embedQuota {
-			d.log.Warn("Discord embed quota reached, not creating more attachments for this notification", "embedQuota", embedQuota)
-			break
-		}
+	_ = images.WithStoredImages(ctx, d.log, d.images,
```
santihernandezc: Why are we not checking the error returned here?
```suggestion
	err := images.WithStoredImages(ctx, d.log, d.images,
```

receivers/discord/discord.go
```diff
@@ -219,66 +217,49 @@ func (d Notifier) SendResolved() bool {
...
-		// We got an attachment, either using the image URL or bytes.
-		attachments = append(attachments, attachment)
-		embedsUsed++
-	}
+			// We got an attachment, either using the image URL or bytes.
+			attachments = append(attachments, *attachment)
+			return nil
+		}, alerts...)
 
 	return attachments
```
santihernandezc, line 244:
```suggestion
        if err != nil {
		d.log.Error("failed to create attachments for Discord", "error", err)
	}
	return attachments
```

Only return the attachments if you're sure it's safe to do so, otherwise return `nil` on error.

receivers/discord/discord_test.go
```diff
@@ -555,18 +565,25 @@ func TestNotify_WithImages(t *testing.T) {
 			UseDiscordUsername: false,
 		}
 
+		tokenStore := images.NewFakeTokenStore(15)
+		imageProvider := images.NewTokenProvider(tokenStore, &logging.FakeLogger{})
+
 		// Create 10 alerts with an image each, Discord's embed limit is 10, and we should be using a maximum of 9 for images.
```
santihernandezc: I don't think this comment is relevant, right?
```suggestion
```

receivers/email/email.go
```diff
@@ -61,18 +58,17 @@ func (en *Notifier) Notify(ctx context.Context, alerts ...*types.Alert) (bool, e
 	embeddedContents := make([]receivers.EmbeddedContent, 0)
 	_ = images.WithStoredImages(ctx, en.log, en.images,
```
santihernandezc: Error checking here too.

santihernandezc (commented):
Nice! Please ping me for a re-review.

images/images.go
```diff
@@ -17,61 +15,30 @@ var (
...
+}
+
 type Image struct {
-	Token     string
-	Path      string
-	URL       string
-	CreatedAt time.Time
+	// URL is the public URL of the image. This URL should not be treated as a trusted source and should not be
+	// downloaded directly. RawData should be used to retrieve the image data.
+	URL string
```
yuri-tceretian, line 31: I wonder if it's better to use a value type here, ie *url.URL which would automatically assume that it's a valid URI?

yuri-tceretian (commented):
LGTM. I like the approach! Let's address the Santiago's comments before merging.

images/providers.go
JacobsonMT: Good point, but I'd like the keep the error generic. I'll add context when used (similar to how we do validation errors). Ex:
```go
fmt.Errorf("%w: URLProvider does not support raw image data", ErrImageUploadNotSupported)
```

images/images.go
JacobsonMT: It's tempting but considering we won't be using any of the features of the `URL` type, I'm not sure it's worth it. Do we even care if the url is not parseable? We'd likely just use the string anyways.

images/utils.go
JacobsonMT: `alert.String()` is already in the context of this logger from `WithStoredImages`

receivers/discord/discord.go
JacobsonMT: Ah, this comes from reverting the code to when it used to use `WithStoredImages`, see https://github.com/grafana/alerting/commit/793c67215ba14d6515b93707caeb7e6d9bf82189#diff-f78cd42a7a3a05c1955ac8b712742062386eaaebb3942dfcd59df91982a9b3d4L208.

It never checked the error from what I can tell, not sure why though. Possibly the intention was to use any attachments that were successful if one failed, but I can check the error and make that explicit instead.

receivers/discord/discord_test.go
JacobsonMT: I'm not sure what you mean, I didn't change the logic of this test and it seems correct?

santihernandezc: I meant that this comment is not accurate. We're creating 15 alerts in this test, not 10.

santihernandezc (approved):
LGTM 🚀

PR #278, opened 2025-02-07
WaitStage for Grafana Alertmanager

yuri-tceretian:
This PR introduces a new implementation of WaitStage that has one difference from the default one:
- log message in the case when the pipeline is resumed. This should give us a better picture of the pipeline timeline while investigating in HA mode.


notify/stages/wait_stage.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/prometheus/alertmanager/notify"
+	"github.com/prometheus/alertmanager/types"
+)
+
+type PeerInfo interface {
+	Position() int
+}
```
titolins, line 15: It feels we're leaking clustering information which should ideally be abstracted from the pipeline. Why not keeping the waitFunc or simplifying it to accept a duration instead?

notify/stages/wait_stage.go
```diff
@@ -0,0 +1,62 @@
...
+		return ctx, alerts, nil
+	}
+	peerPosition := ws.peer.Position()
+	wait := time.Duration(peerPosition) * ws.timeout
+	if wait == 0 {
+		return ctx, alerts, nil
+	}
+
+	t := time.NewTimer(wait)
+	defer t.Stop()
```
titolins: Didn't know about this, good catch. It seems like this was fixed in 1.23 though. As per the `time.After` [docs](https://pkg.go.dev/time#After):

> Before Go 1.23, this documentation warned that the underlying [Timer](https://pkg.go.dev/time#Timer) would not be recovered by the garbage collector until the timer fired, and that if efficiency was a concern, code should use NewTimer instead and call [Timer.Stop](https://pkg.go.dev/time#Timer.Stop) if the timer is no longer needed. As of Go 1.23, the garbage collector can recover unreferenced, unstopped timers. There is no reason to prefer NewTimer when After will do.

Considering Grafana is already in 1.23, we could consider updating alerting and get this for free. We could keep this for the logs in any case - wdyt?

notify/stages/wait_stage.go
yuri-tceretian: 1. We leak it already in other steps such as MeshStage. 
2. The position of the cluster in the ring changes over time and I think it would be beneficial to log the position when the stage was executed.

notify/stages/wait_stage.go
yuri-tceretian: Good catch :) can change it back then

titolins (approved):
🚀

PR #280, opened 2025-02-11
Add Jira integration

santihernandezc:
This PR introduces support for JIRA integration in Grafana Managed Alertmanager. 
It's almost carbon copy of the https://github.com/prometheus/alertmanager/pull/3590 (credits to @jkroepke) with couple changes:
- [`b52aeaa` (#280)](https://github.com/grafana/alerting/pull/280/commits/b52aeaa50aff5f37dc8b37608e16a6476b077539) Configuration extended with option `dedup_key_field` that allows to specify a custom field to store deduplication key. The field must be a text field. If it is not specified, the deduplication key is added to labels.
- [`1578f74` (#280)](https://github.com/grafana/alerting/pull/280/commits/1578f747ca1baedaf10ab8a3b4ec263ac7bc4a67) changes how description is handled in v3 API: 
    - if the payload is not a valid JSON it treats it as string, truncates it and created and ADF document
    - if payload is a valid JSON, it does not truncate it and puts into the body as is. The code in the upstream PR would fail to parse truncated JSON. This way it will be on API to reject it as too long. 
- [`d9b37ab` (#280)](https://github.com/grafana/alerting/pull/280/commits/d9b37ab7a63aa99d715d5fe71f13cb4835daf365) changes how empty transition settings are treated. In contrast to the upstream implementation that looks for empty-string transition and fails notification, this integration skips resolving\reopening of integration. 
- to match the behavior of other integrations, errors in expanding template do not cause the integration to fail. Instead, it follows the Grafana's paradigm of best effort to notify user even if the payload is not what user expects and falls back to the default templates.
- handling of field `Fields`. 
   - The original PR does conversion of `map[any]any` to `map[string]any` [here ](https://github.com/prometheus/alertmanager/pull/3590/files#diff-c72db7eea319d89bf8b0a8cc52233489f60b917bd9765caaff49baf639338188R146) because yaml unmarshaller produced result that is not supported by json marshaller. In the case of Grafana the integration configuration is stored in JSON and therefore, it is not affected by this problem. 
   - However, due to limitations put by Grafana's UI framework for integrations, the Fields could be only a map[string]string. To workaround this limitation the config parser is [extended](https://github.com/grafana/alerting/blob/09b057fdde0fe14f0ff5125d13e5ee482109a236/receivers/jira/config.go#L127-L146) to scan values and parse if string is a valid JSON
 
### How integration works
[Diagram](https://www.plantuml.com/plantuml/uml/bP71JW8n48RlVOevmi4NO0y6KGmQDG6Dvx8xoquedKrd5jotjor208c9FJHjvf__-pkLDVdM1hk1fItNgy2ThoRiE806woMMqAkHeygMIblI_fxiXdrrTz7qsP3YOiDhD4P97SBJR3w6LrQg2GFSx8yzlA0sNC7KySuOKHlKcD42RIWRUFWYKV8hc8bq2CiU7YEYXMVi2sEeXi4q4M-7e0rw6FGeGmDnZJxQoYeU3Gg3JX06dlVrr3fs61GMQhKJ884v2hijNd5BIvJRV3atoh19m0Jpk8i-CMXnfSjNL9lzdeFRH1UeASHubyWao4uf5xVeVwVACP9kZaa09VkQLbsubd_qddwGbDY3Chn5VV5ZOOxHJfvrXYyTov_uFBN_y2VPyP3W03sTK1hRDm00)
diagram-733674009971547440

### Notes for reviewer
Due to some security reason, the Grafana's webhook sender does not support GET requests. I initiated extraction of it in https://github.com/grafana/alerting/pull/281 to alerting module to be able to change it and have more control over its behavior. In the meantime, I worked around GET requests [`449c192` (#280)](https://github.com/grafana/alerting/pull/280/commits/449c192b202f20e7b60ca132f2ef307549475218). I will remove once the migrated sender will be able to process GET requests.


jkroepke:
Some changes are great. I hope
Some of them can be done upstream

http/webhook.go
```diff
@@ -0,0 +1,87 @@
...
+	_, err = url.Parse(cmd.URL)
+	if err != nil {
+		// Should not be possible - NewRequestWithContext should also err if the URL is bad.
+		return err
+	}
+
+	request.Header.Set("User-Agent", "Grafana")
+
+	if cmd.User != "" && cmd.Password != "" {
+		request.Header.Set("Authorization", GetBasicAuthHeader(cmd.User, cmd.Password))
```
santihernandezc: There's no need to create a function for this, Go already has one in its stdlib.

```suggestion
		request.SetBasicAuth(cmd.User, cmd.Password)
```

http/webhook.go
```diff
@@ -0,0 +1,87 @@
...
+	for k, v := range cmd.HTTPHeader {
+		request.Header.Set(k, v)
+	}
+
+	resp, err := receivers.NewTLSClient(cmd.TLSConfig).Do(request)
+	if err != nil {
+		return redactURL(err)
+	}
+	defer func() {
+		_ = resp.Body.Close()
```
santihernandezc, line 51: Not strictly necessary, feel free to ignore, but wdyt about passing a logger and logging this error if it ever occurs?

receivers/jira/jira.go
```diff
@@ -0,0 +1,376 @@
...
+		tmpl:    template,
+		conf:    cfg,
+		retrier: &notify.Retrier{RetryCodes: []int{http.StatusTooManyRequests}},
+	}
+}
+
+// Notify implements the Notifier interface.
+func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
+	if n.conf.URL == nil {
+		// this should not happen but it's better to avoid panic
```
santihernandezc: Nit
```suggestion
		// This should not happen but it's better to avoid panic.
```

receivers/jira/jira.go
```diff
@@ -0,0 +1,376 @@
...
+		return false, fmt.Errorf("missing JIRA URL")
+	}
+	var (
+		alerts = types.Alerts(as...)
+	)
+
```
santihernandezc: Why not just...
```suggestion

	alerts := types.Alerts(as...)
```

receivers/jira/jira.go
```diff
@@ -0,0 +1,376 @@
...
+			return false, nil
+		}
+		path = "issue"
+		method = http.MethodPost
+		logger.Debug("create new issue")
+	} else {
+		path = "issue/" + url.PathEscape(existingIssue.Key)
+		method = http.MethodPut
+		logger.Debug("updating existing issue", "issue_key", existingIssue.Key)
+	}
```
santihernandezc: Just a suggestion, you can ignore the `method` and `path` changes, but please change "create" to "creating" in the debug log line.
```suggestion
	method := http.MethodPost
	path := "issue"
	if existingIssue == nil {
		// Do not create new issues for resolved alerts.
		if alerts.Status() == model.AlertResolved {
			return false, nil
		}
		logger.Debug("creating new issue")
	} else {
		path = "issue/" + url.PathEscape(existingIssue.Key)
		method = http.MethodPut
		logger.Debug("updating existing issue", "issue_key", existingIssue.Key)
	}
```

receivers/jira/jira.go
```diff
@@ -0,0 +1,376 @@
...
+
+	for i, label := range n.conf.Labels {
+		label = strings.TrimSpace(renderOrDefault(fmt.Sprintf("labels[%d]", i), label, ""))
+		if label == "" {
+			continue
+		}
+		fields.Labels = append(fields.Labels, label)
+	}
+	sort.Strings(fields.Labels)
```
santihernandezc: Nit: From the [docs](https://pkg.go.dev/sort#Strings), calling `sort.Strings()` in Go 1.22 or higher simply calls `slices.Sort()`.
```suggestion
	slices.Sort(fields.Labels)
```

receivers/jira/jira.go
```diff
@@ -0,0 +1,376 @@
...
+		fields.Priority = &idNameValue{Name: priority}
+	}
+
+	if n.conf.DedupKeyFieldName != "" {
+		if fields.Fields == nil {
+			fields.Fields = make(map[string]any)
+		}
+		fields.Fields[fmt.Sprintf("customfield_%s", n.conf.DedupKeyFieldName)] = groupID
+	} else {
+		// this label is added to be able to search for an existing one
```
santihernandezc:
```suggestion
		// This label is added to be able to search for an existing one.
```

receivers/jira/jira.go
```diff
@@ -0,0 +1,376 @@
...
+		fields.Labels = append(fields.Labels, fmt.Sprintf("ALERT{%s}", groupID))
+	}
+
+	return issue{Fields: fields}
+}
+
+func (n *Notifier) prepareDescription(desc string, logger logging.Logger) any {
+	if strings.HasSuffix(strings.TrimRight(n.conf.URL.Path, "/"), "/3") {
+		// V3 API supports structured description in ADF format
+		// check if the payload is a valid JSON and assign it in that case
```
santihernandezc:
```suggestion
		// V3 API supports structured description in ADF format.
		// Check if the payload is a valid JSON and assign it in that case.
```

receivers/jira/jira.go
```diff
@@ -0,0 +1,376 @@
...
+		if json.Valid([]byte(desc)) {
+			// we do not check the size of description if it's structured data because we can't truncate it.
```
santihernandezc, line 182:
```suggestion
			// We do not check the size of the description if it's structured data because we can't truncate it.
```

receivers/jira/jira.go
```diff
@@ -0,0 +1,376 @@
...
+			var issueDescription any
+			err := json.Unmarshal([]byte(desc), &issueDescription)
+			if err == nil {
+				return issueDescription
+			}
+			logger.Warn("Failed to parse description as JSON. Fallback to string mode", "error", err)
+		}
+
+		maxLen := MaxDescriptionLenRunes - adfDocOverhead
+		// if it's just a text. Create a document. Consider the document overhead while truncating
```
santihernandezc:
```suggestion
		// If it's just text, create a document.
		// Consider the document overhead while truncating.
		maxLen := MaxDescriptionLenRunes - adfDocOverhead
```

receivers/jira/jira_test.go
```diff
@@ -0,0 +1,835 @@
...
+
+			// Execute
+			desc := notifier.prepareDescription(tt.description, logger)
+			assert.EqualValues(t, tt.expected, desc)
+		})
+	}
+}
+
+func TestPrepareIssueRequestBody(t *testing.T) {
+
```
santihernandezc:
```suggestion
```

santihernandezc (commented):
Looks good!
I left a few comments. Also, please consider capitalizing log lines.

receivers/jira/jira.go
yuri-tceretian: I did this intentionally. I think it more readable this way when you do actions in if branches. Anyway, I am fine with this change

yuri-tceretian (approved):

PR #284, opened 2025-02-18
Alerting: Fix token-based Slack image upload to work with channel names

JacobsonMT:
Fixes `channel_not_found` error when finalizing an image upload for a Slack integration that uses both `Token` and channel `Name` (`#example`) as the recipient instead of channel `ID` (`C123ABC456`).

The cause was a previous [change](https://github.com/grafana/alerting/pull/256) to support `files_upload_v2` that required the use of the [files.completUploadExternal](https://api.slack.com/methods/files.completeUploadExternal) endpoint which does not support referencing channels by name, only ID.

Now we obtain the channel ID from the `chat.postMessage` response specifically for these bot token image uploads.

### DIFF
--- a/receivers/slack/slack.go
+++ b/receivers/slack/slack.go
@@ -14,6 +14,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


receivers/slack/slack.go
```diff
@@ -517,16 +525,15 @@ func initialCommentForImage(alert *types.Alert) string {
...
-		sb.WriteString(string(k))
-		sb.WriteString(" = ")
-		sb.WriteString(string(v))
-		if n < len(alert.Labels)-1 {
-			sb.WriteString(", ")
-			n++
+	if len(alert.Labels) == 0 {
+		sb.WriteString("None")
+	} else {
+		lstrs := make([]string, 0, len(alert.Labels))
```
JacobsonMT, line 531: Needed to change this as it's not deterministic and was causing flaky tests.

NOTE: The logic here is incorrect IMO as it's using all labels including private ones. I will fix this in a [followup](https://github.com/grafana/alerting/pull/286) to use `templates.ExtendedAlert` which has sanitized labels.

Ex:

`
Firing: always firing prom (copy), Labels: __grafana_autogenerated__ = true, __grafana_receiver__ = Both Slack Types, __grafana_route_settings_hash__ = 554a8065f06504ff, alertname = always firing prom (copy), __alert_rule_uid__ = cedgbhztxup6oe, prom = true, grafana_folder = Grafana Common
`

yuri-tceretian (approved):
LGTM

PR #285, opened 2025-02-18
Support HMAC SHA256 request signing in the http client

alexander-akhmetov:
Adds optional HMAC SHA256 signing to the HTTP client.

A new HMACRoundTripper (in http/hmac.go) computes and adds a signature for outgoing webhooks. The signature is computed from the request body.

Optionally, a timestamp header name can be specified, and if it's set, the signature is computed from `{timestamp}:{body}`, and the timestamp is added to the headers for the client to verify.

Other implementation examples:
- [Slack](https://api.slack.com/authentication/verifying-requests-from-slack) (also adds a timestamp)
- [Docusign](https://developers.docusign.com/platform/webhooks/connect/setting-up-hmac/)

Draft PR to Grafana: https://github.com/grafana/grafana/pull/100960 if you want to test it locally.

Part of https://github.com/grafana/alerting-squad/issues/1040

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -10,6 +10,8 @@ import (
 	"net/http"
 	"net/url"
 
+	"github.com/benbjohnson/clock"
+
 	"github.com/grafana/alerting/logging"
 	"github.com/grafana/alerting/receivers"
 )
@@ -37,7 +39,7 @@ func NewClient(log logging.Logger, cfg ClientConfiguration) *Client {
 }
 
 func (ns *Client) SendWebhook(ctx context.Context, webhook *receivers.SendWebhookSettings) error {
-	// This method is carbon copy of https://github.com/grafana/grafana/blob/71d04a326be9578e2d678f23c1efa61768e0541f/pkg/services/notifications/webhook.go#L38
+	// This method was moved from https://github.com/grafana/grafana/blob/71d04a326be9578e2d678f23c1efa61768e0541f/pkg/services/notifications/webhook.go#L38
 	if webhook.HTTPMethod == "" {
 		webhook.HTTPMethod = http.MethodPost
 	}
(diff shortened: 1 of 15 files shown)


http/client.go
```diff
@@ -72,7 +74,24 @@ func (ns *Client) SendWebhook(ctx context.Context, webhook *receivers.SendWebhoo
 		request.Header.Set(k, v)
 	}
 
-	resp, err := receivers.NewTLSClient(webhook.TLSConfig).Do(request)
+	client := receivers.NewTLSClient(webhook.TLSConfig)
```
titolins: It might make sense changing `NewTLSClient` to something like `NewAuthClient` and merging HMAC/TLS logic. WDYT?

http/client.go
```diff
@@ -72,7 +74,24 @@ func (ns *Client) SendWebhook(ctx context.Context, webhook *receivers.SendWebhoo
...
+
+	if webhook.HMACConfig != nil {
+		ns.log.Debug("Adding HMAC roundtripper to client")
+		client.Transport, err = NewHMACRoundTripper(
```
titolins, line 81: Why HMAC > TLS?

http/hmac.go
```diff
@@ -0,0 +1,92 @@
...
+	header          string
+	timestampHeader string
+}
+
+// NewHMACRoundTripper creates a new HMACRoundTripper that wraps the provided RoundTripper.
+// It signs requests using the provided secret key and places the signature in the specified header.
+// If header is empty, it defaults to "X-Grafana-Alert-Signature".
+// If timestampHeader is non-empty, the current timestamp will be included in the signature
+// calculation and set in the specified header.
+func NewHMACRoundTripper(wrapped http.RoundTripper, clk clock.Clock, secret, header, timestampHeader string) (*HMACRoundTripper, error) {
```
titolins, line 40: Just wondering why I've pushed the TLS client to the receivers package. I'd say it makes sense to have both of them in the same place. Wdyt?

titolins (commented):
Just a couple of nits

http/client.go
alexander-akhmetov: It's wrapping the TLS transport, so should be fine?

http/hmac.go
alexander-akhmetov: I think it was just before we copied the http client in https://github.com/grafana/alerting/pull/281 :) Probably now it makes sense to have them here

http/client.go
titolins: Yep, my bad - missed this 👍

http/hmac.go
alexander-akhmetov: Moved it :) https://github.com/grafana/alerting/pull/285/commits/d0aa3e899242ad5eab291eb8e2e4ec2e7f78c09e

http/client.go
alexander-akhmetov: Discussed in Slack, agreed to keep it as it is for now

titolins (approved):
🚀

PR #286, opened 2025-02-18
Alerting: Sanitize Slack image upload comment labels

JacobsonMT:
See https://github.com/grafana/alerting/pull/284#discussion_r1960647474

The logic for Slack image upload comment creation is using all labels, including private ones. This sanitizes the label by using `templates.ExtendedAlert` as the input which already includes sanitized data ready for templating.

Based on https://github.com/grafana/alerting/pull/284, so that needs to merge first.

### Old Example:
`Firing: always firing prom (copy), Labels: __grafana_autogenerated__=true, __grafana_receiver__=Both Slack Types, __grafana_route_settings_hash__=554a8065f06504ff, alertname=always firing prom (copy), __alert_rule_uid__=cedgbhztxup6oe, prom=true, grafana_folder=Grafana Common`

### New Example:
`Firing: always firing prom (copy), Labels: alertname=always firing prom (copy), prom=true, grafana_folder=Grafana Common`


receivers/slack/slack.go
```diff
@@ -511,29 +514,26 @@ func (sn *Notifier) SendResolved() bool {
...
 		sb.WriteString("None")
 	} else {
-		lstrs := make([]string, 0, len(alert.Labels))
-		for l, v := range alert.Labels {
-			lstrs = append(lstrs, fmt.Sprintf("%s=%s", l, v))
-		}
-		sort.Strings(lstrs)
-		sb.WriteString(strings.Join(lstrs, ", "))
+		sb.WriteString(alert.Labels.SortedPairs().String())
 	}
```
yuri-tceretian, line 538: It looks like we will put label `alertname` twice: first time - above, and the second time - as a label. 
Btw, SortedPairs has a special logic that always sets label `alertname` as the first element so you can just do  
```go
alert.Labels.SortedPairs()[1:].String()
```

yuri-tceretian (approved):
One comment. Otherwise, LGTM

PR #290, opened 2025-02-24
Try additional scopes when testing templates with root scope fails

JacobsonMT:
Template testing currently tests all definitions using the `ExtendedData` as the scope (`.`). However, some template definitions may be intended to be used with other scopes, such as `.Alerts` or a single `ExtendedAlert`. 

This limitation is somewhat mitigated by the fact that we restrict which definitions within a template are returned as results during testing. We only return "top-level" definitions, i.e. definitions that aren't reference by other definitions in the template, and definitions meant for non-root scopes are *usually* not top-level.

One clear counter-example is the `pagerduty.default.instances` default definition. Which is both top-level and meant to be used with the `.Alerts` scope. This causes a persistent error to display when preview the default definitions:

image

This leaves us with a few options:

1. Ask for intended scope when testing templates.
2. Analyze the definition to attempt to determine which scope it's intended for.
3. Attempt a pre-determined set of more specific scope if the root scope fails.

Thoughts:

1. Is tempting but is a larger change that has rippling effects on the Template model and UI/UX. In the future though, if we end up separating the `Template` model to be more granular in k8s API, this could become an good option.
2. This is likely a rabbit-hole of brittleness.
3. Fairly safe incremental improvement that, while not covering all possible scopes, is extensible and simple. Performance will be slower but the endpoint is not high volume and the performance impact only effects failing templates.

This PR implements 3.

Additional Scopes:
- `.Alerts`
- `.Alert` (aka .Alerts[0])

### DIFF
--- a/notify/grafana_alertmanager.go
+++ b/notify/grafana_alertmanager.go
@@ -1,7 +1,6 @@
 package notify
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


fayzal-g (approved):
lgtm!

PR #300, opened 2025-03-06
Sync flush time

titolins:
## Context
The idea is to have all AM instances flushing alerts at the same/similar time. This should reduce the number of duplicated notifications and help reduce escalations.

The strategy is to add a new stage to the notification pipeline. On https://github.com/grafana/prometheus-alertmanager/pull/101, we're adding the pipeline time to the notification entry, so we can retrieve this in the next flush. We then compare the current pipeline time with the expected next flush (summing the previous pipeline with the group wait and considering a margin). If the current execution time is similar to the expected next flush, we execute it immediately. Otherwise, wait for `nextFlush - currentPipelineTime`.

## Depends on
- https://github.com/grafana/prometheus-alertmanager/pull/101

## Related PR's
- https://github.com/grafana/grafana/pull/101931


go.mod
```diff
@@ -105,7 +105,7 @@ require (
 )
 
 // Using a fork of the Alertmanager with Alerting Squad specific changes.
-replace github.com/prometheus/alertmanager => github.com/grafana/prometheus-alertmanager v0.25.1-0.20250305143719-fa9fa7096626
+replace github.com/prometheus/alertmanager => github.com/grafana/prometheus-alertmanager v0.25.1-0.20250306181800-10368d39a559
```
titolins: should be updated on the AM PR is merged

go.sum
```diff
@@ -194,8 +194,8 @@ github.com/google/uuid v1.5.0 h1:1p67kYwdtXjb0gL0BPiP1Av9wiZPo5A8z2cWkTZ+eyU=
 github.com/google/uuid v1.5.0/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
 github.com/googleapis/gax-go/v2 v2.0.4/go.mod h1:0Wqv26UfaUD9n4G6kQubkQ+KchISgw+vpHVxEJEs9eg=
 github.com/googleapis/gax-go/v2 v2.0.5/go.mod h1:DWXyrwAJ9X0FpwwEdw+IPEYBICEFu5mhpdKc/us6bOk=
-github.com/grafana/prometheus-alertmanager v0.25.1-0.20250305143719-fa9fa7096626 h1:QsMYtDseSPq8hXvoNtA64unFiawJaE5kryizcMsVZWg=
-github.com/grafana/prometheus-alertmanager v0.25.1-0.20250305143719-fa9fa7096626/go.mod h1:FGdGvhI40Dq+CTQaSzK9evuve774cgOUdGfVO04OXkw=
+github.com/grafana/prometheus-alertmanager v0.25.1-0.20250306181800-10368d39a559 h1:E7az+c68g6E5X0mt8ZRm9+wZ530EWSQUTHgVZ1AKBRQ=
```
titolins: likewise

notify/grafana_alertmanager.go
```diff
@@ -178,6 +182,10 @@ type GrafanaAlertmanagerConfig struct {
 
 	// PipelineAndStateTimestampsMismatchAction defines the action to take when there's a mismatch in pipeline and state timestamps.
 	PipelineAndStateTimestampsMismatchAction stages.Action
+	// SyncFlushAction defines weather to run the sync flush stage and in what mode.
```
rwwiv:
```suggestion
	// SyncFlushAction defines whether to run the sync flush stage and in what mode.
```

notify/stages/sync_flush_stage_test.go
```diff
@@ -0,0 +1,266 @@
...
+		skipNow        bool
+		skipGroupWait  bool
+		expectedErr    bool
+	}{
+		{
+			name:         "no entries",
+			sync:         true,
+			entries:      []*nflogpb.Entry{},
+			pipelineTime: now,
+			expectedErr:  false,
```
rwwiv: WDYT about actually including the expected error here, or at least a string we can check against?

notify/stages/sync_flush_stage.go
```diff
@@ -0,0 +1,125 @@
...
+		"receiver", sfs.recv.GroupName,
+		"integration", sfs.recv.Integration,
+		"entry_pipeline_time", entry.PipelineTime,
+		"wait", wait,
+		"sync", sfs.sync,
+	)
+
+	if sfs.sync {
+		select {
+		case <-time.After(wait):
```
rwwiv, line 131: I wonder if we should update the context with a new "Now" based on the wait.

notify/stages/sync_flush_stage_test.go
titolins: yeah, I got a bit lazy on these tests tbh 🙈 
added proper errors now and comparing against those 👍

notify/stages/sync_flush_stage.go
titolins: That's a good point, maybe we should.. not sure what would be the impact of that in the other stages, but if we're simulating a flush delay I guess that it would be the right thing..

cc @yuri-tceretian since it would affect the extra dedup specifically

yuri-tceretian: No, the timeNow must be the time when snapshot of the aggrGroup was taken. This step runs after that and does not make the new snapshot. 
I think after it wakes up it needs to run some kind of "coordination step" but without warning log and proceed if there wasn't any state from the future. Otherwise, it should immediately exit pipeline and start waiting for group_interval

notify/stages/sync_flush_stage.go
```diff
@@ -0,0 +1,139 @@
...
+	if !ok {
+		return ctx, nil, ErrMissingNow
+	}
+
+	groupInterval, ok := notify.GroupInterval(ctx)
+	if !ok {
+		return ctx, nil, ErrMissingGroupInterval
+	}
+
+	entries, err := sfs.nflog.Query(nflog.QGroupKey(gkey), nflog.QReceiver(sfs.recv))
```
yuri-tceretian, line 99: I think this will break when user adds a new integration to the receiver, and basically, will delay old integrations while not delaying new integration.

notify/grafana_alertmanager.go
```diff
@@ -892,6 +901,9 @@ func (am *GrafanaAlertmanager) createReceiverStage(name string, integrations []*
 			Idx:         uint32(integrations[i].Index()),
 		}
 		var s notify.MultiStage
+		if stage := stages.NewSyncFlushStage(notificationLog, recv, syncAct, syncMargin); stage != nil {
```
yuri-tceretian, line 904: I think this stage should run before Fanout stage, perhaps right after `meshStage` or right after inhibitionStage . This will guarantee that the pipeline is delayed correctly.

Issue #134, opened 2023-11-04
Webhook Receiver should support custom headers


njohnstone2:
The webhook receiver should have the ability to add arbitrary headers to a http request.

When configuring a webhook contact point it is common that the webhook receiver requires additional arbitrary headers to process the request (e.g. "X-" headers).

This issue would also require a UI change in grafana to allow the additional headers to be defined in the UI

armandgrillet:
The challenge is that we want to unify our receivers with the ones in the Prometheus Alertmanager repository in the long term, and this issue + PR go against that. You should propose the idea to https://github.com/prometheus/alertmanager/issues instead.

armandgrillet:
Related PRs:

- https://github.com/grafana/alerting/pull/135
- https://github.com/grafana/grafana/pull/77668

This does not change the fact that this should first be done in the Prometheus Alertmanager repository so that we do not go against upstream.

WesselAtWork:
Turns out the intended way of dealing with arbitrary alerts in prom/alertmanager is to use a custom middleware that forwards the webhook to the upstream.

There seems to be many many feature requests the effect of "Please Add X Integration" or "Please modify webhooks so we can make arbitrary requests" or a more reasonable "Please allow us to template the body of the webhooks"

It looks like they have decided (implicitly) that the burdun of implementation is on the server side.

I doubt they will be open to adding custom headers, even if they were static.
https://github.com/prometheus/alertmanager/issues/2268

See also:
https://github.com/prometheus/alertmanager/pull/709#issuecomment-312595545
https://github.com/prometheus/alertmanager/pull/2184#issuecomment-586647546
https://github.com/prometheus/alertmanager/issues/1496#issuecomment-409930740


---
Are there plans to add additional grafana-only receivers?  
I don't think there are connector plugins available but I could be wrong.

christianherweg0807:
That would be great, especially for routing messages, when replacing the deprecated teams Webhook with PowerAutomate Workflows for multiple Teams Channels.

We think about headers like ChannelID & TeamID to route the Alarm to the correct destination & team

JacobsonMT:
Fixed with https://github.com/grafana/alerting/pull/309

PR #309, opened 2025-04-09
Webhook Improvements - Templateable Payloads

JacobsonMT:
## Overview
This PR introduces enhancements to the webhook notifier, focusing on three main areas:
1. Customizable webhook payloads using templates
2. Support for custom HTTP headers
3. New template functions to provide support for creating well-formatted, valid JSON in payloads.

### 1. Custom Payload Templates
- Added support for custom JSON payload templates in webhook notifications
- Users can now define their own payload structure using templates
- Includes a new `CustomPayload` field with:
  - `Template` field for defining the payload structure
  - `Vars` field for providing additional template variables accessible in context via `.Vars.<variable_name>`
- Provides an example payload template which is equivalent to the default payload format for demonstration and regression testing

### 2. Custom HTTP Headers
- Added support for custom HTTP headers in webhook requests, example usage would be to set the `Content-Type` header for your custom payload.
- Implemented security measures to prevent misuse:
  - Restricted headers list to prevent security issues
  - Headers like `Authorization`, `Cookie`, `Host` are protected

### 3. New Template Functions
- Introduced new template functions inspired by [gomplate](https://docs.gomplate.ca/):
  - `coll` namespace for collection operations (`Dict`, `Slice`, `Append`)
  - `data` namespace for JSON operations (`JSON`, `ToJSON`, `ToJSONPretty`)
  - `tmpl` namespace for template operations compatible with pipelines (`Inline`, `Exec`)
  - `time` namespace for time operations (`Now`)
- These functions enable more complex and flexible payload formatting, especially when creating JSON.
- Currently gomplate does not allow modularized use of select functions or namespaces, so importing it directly would require a significant increase in dependencies. In addition, certain necessary functions have extended functionality in gomplate that makes them risky to use as-is such as the `eJSON` decryption capability of `data.JSON` which reads from environment variables. That being said, none of this is insurmountable, and the plan should be to eventually incorporate the gomplate library directly and more broadly.

### 4. Related Improvements
- Added new fields to `ExtendedData` and `ExtendedAlert` to make sure custom payloads have access to all of the information the default webhook payload has:
  - `ExtendedAlert.OrgId`
  - `ExtendedData.GroupKey`
  - `ExtendedData.TruncatedAlerts`

## Related Requests

Issues for supporting new notifiers in Grafana are fairly common and are generally reasonable requests. However, they can come with significant burden of maintenance. Some examples:
- [Telegram Local Bot API support](https://github.com/grafana/grafana/issues/81534)
- [Alerting: Extend Pagerduty notification for Openduty](https://github.com/grafana/grafana/issues/6504)
- [Feature request: New contact point type: Amazon EventBridge](https://github.com/grafana/grafana/issues/82645)
- [FR: Alerting Contact Point Matrix](https://github.com/grafana/grafana/issues/82764)
- [Add web push notifications to Grafana Alerting](https://github.com/grafana/grafana/issues/10306)

Requests for related features are some of the most popular in Alerting:
- [[Feature request] Custom json passed as a webhook alert](https://github.com/grafana/grafana/issues/6956) 
- [Support Backend plugins for Alert notifiers](https://github.com/grafana/grafana/issues/16004)
- [Custom Headers on HTTP webhook](https://github.com/grafana/grafana/issues/9813#top)
- [Support custom headers on the webhook contact point](https://github.com/grafana/grafana/issues/50558#top)
- [Webhook Receiver should support custom headers](https://github.com/grafana/alerting/issues/134#top)

Upstream Prometheus Alertmanager also has significant community requests, although consensus so far has been to keep the webhook notifier as-is with recommendations to set up a small web server to act as a webhook bridge:
- [it's more useful for webhook_config support template and custom the header](https://github.com/prometheus/alertmanager/issues/2268#top)
- [Added functionality to alter Webhook message](https://github.com/prometheus/alertmanager/pull/1422)
- [Reintroduce AWS SNS notifications](https://github.com/prometheus/alertmanager/pull/709#issuecomment-312595545)
- [Add templated text support for webhook receivers](https://github.com/prometheus/alertmanager/pull/2184#issuecomment-586647546)
- [Templating support for body of webhook_config](https://github.com/prometheus/alertmanager/issues/1496#issuecomment-409930740)
- [webhook custom style](https://github.com/prometheus/alertmanager/issues/3110#top)

## Future Work

- Import [gomplate](https://docs.gomplate.ca/) and expand template support to more namespaces.

## Custom Payload Examples

Could be as simple as writing JSON directly:
```json
{
  "alert_name": "{{ .CommonLabels.alertname }}",
  "status": "{{ .Status }}",
  "environment": "{{ .Vars.environment }}",
  "custom_field": "{{ .Vars.custom_field }}"
}
```
Or using new functions to help craft valid JSON:
```blade
  {{ coll.Dict 
  "receiver" .Receiver
  "status" .Status
  "alerts" .Alerts
  "groupLabels" .GroupLabels
  "commonLabels" .CommonLabels
  "commonAnnotations" .CommonAnnotations
  "externalURL" .ExternalURL
  "version" "1"
  "orgId"  (index .Alerts 0).OrgId
  "truncatedAlerts"  .TruncatedAlerts
  "groupKey" .GroupKey
  "state"  (tmpl.Inline "{{ if eq .Status \"resolved\" }}ok{{ else }}alerting{{ end }}" . )
  "title" (tmpl.Exec "default.title" . )
  "message" (tmpl.Exec "default.message" . )
  | data.ToJSONPretty " "}}
```

## Review Tips

- New gomplate-based functions are nearly 1:1 with the library, so testing was skipped given we're going to eventually remove these in favour of and import. Any differences are pointed out in comments.
- Grafana-side PR will be easier to test with as it will include UI support and enhancements to the template builder to help when creating JSON.

### DIFF
--- a/receivers/webhook/config.go
+++ b/receivers/webhook/config.go
@@ -18,7 +18,7 @@ type CustomPayload struct {
 
 	// Vars is a map of variables that can be used in the payload template. This is useful for providing
 	// additional context to the payload template without storing it in the template itself.
-	// Variables are accessible in the template through `.Extra.Vars.<key>`.
+	// Variables are accessible in the template through `.Vars.<key>`.
 	Vars map[string]string `json:"vars,omitempty" yaml:"vars,omitempty"`
 }
 
--- a/receivers/webhook/config.go
+++ b/receivers/webhook/config.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"sort"
 	"strconv"
 
 	"github.com/grafana/alerting/receivers"
@@ -29,8 +30,9 @@ type Config struct {
 	AuthorizationScheme      string
 	AuthorizationCredentials string
 	// HTTP Basic Authentication.
-	User     string
-	Password string
+	User         string
+	Password     string
+	ExtraHeaders map[string]string
 
 	Title      string
 	Message    string
@@ -54,6 +56,7 @@ func NewConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Confi
 		Message                  string                   `json:"message,omitempty" yaml:"message,omitempty"`
 		TLSConfig                *receivers.TLSConfig     `json:"tlsConfig,omitempty" yaml:"tlsConfig,omitempty"`
 		HMACConfig               *receivers.HMACConfig    `json:"hmacConfig,omitempty" yaml:"hmacConfig,omitempty"`
+		ExtraHeaders             map[string]string        `json:"headers,omitempty" yaml:"headers,omitempty"`
 
 		Payload *CustomPayload `json:"payload,omitempty" yaml:"payload,omitempty"`
 	}{}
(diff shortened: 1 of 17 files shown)


moustafab (approved):
Looks great! Nice work!

My only thoughts on the implementation are around further improvements in the supported functions. If we anticipate significant churn here it may be good to establish an experimental namespace which we document to allow for a bit of safer additions. I'm hesitant to support to much without clear needs and I think this approach you've taken so far is far preferable to pulling in an entire library until clear need is well documented/established.

Also, templating header values might come in handy in the future but maybe we can wait for a request from the community before adding that.

receivers/webhook/webhook.go
```diff
@@ -66,6 +66,14 @@ func (wn *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error
...
 	var tmplErr error
 	tmpl, data := templates.TmplText(ctx, wn.tmpl, as, wn.log, &tmplErr)
+	data.TruncatedAlerts = &numTruncated
+
+	// Fail early if we can't template the URL.
+	parsedURL := tmpl(wn.settings.URL)
+	if tmplErr != nil {
+		return false, tmplErr
+	}
+	tmplErr = nil // Reset the error for the next template.
```
titolins: nit: this shouldn't be required, as cases where `tmplErr` is not nil will be caught by the if above. it might help make some sense of this implementation though 😅

templates/gomplate/data.go
```diff
@@ -0,0 +1,87 @@
+package gomplate
```
titolins, line 28: mostly curious about such cases, but should we include gomplate's license?

receivers/webhook/config.go
```diff
@@ -99,5 +121,61 @@ func NewConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Confi
...
+// Returns the new headers and a list of omitted headers.
+func OmitRestrictedHeaders(headers map[string]string) (map[string]string, []string) {
+	safeHeaders := make(map[string]string, len(headers))
+	omitted := make([]string, 0)
+	for k, v := range headers {
+		if _, ok := restrictedHeaders[http.CanonicalHeaderKey(k)]; ok {
+			omitted = append(omitted, k)
+			continue
+		}
+		safeHeaders[http.CanonicalHeaderKey(k)] = v
```
titolins: I do like the idea of standardizing the headers, but it could prove troublesome for users. Sometimes users might not have complete control over the receiver's implementation and that might make use of non-"canonicalized" headers.. I've had issues using [header.Add](https://pkg.go.dev/net/http#Header.Add) before because of that.
It still makes sense using it for comparison, but maybe it's worth keeping the safe headers in the original format. WDYT?

titolins (approved):
weekly board maintenance ✅ 😂 
great work 👏

receivers/webhook/config.go
JacobsonMT: :+1: This is a very good point and 100% correct IMO. I'll keep them as-is.

templates/gomplate/data.go
JacobsonMT: Good question, I'm not sure. I'll ask around.

JacobsonMT: Added their MIT license to the individual gomplate files evolved from https://github.com/hairyhenderson/gomplate

PR #314, opened 2025-04-28
chore: address actions lints

moustafab:
Reduce permissions to specific needs


JacobsonMT (approved):

dannyc-grafana (approved):
These look to be reasonable fixes.

PR #317, opened 2025-04-30
Refactor receiver webhook client to support custom dialers

JacobsonMT:
- Introduced client options in HTTP client creation similar to upstream.
- Added support for custom `DialContext` in webhook client.
- Enhanced integration tests to validate custom dialer behavior.

Part of a series of PRs for full support. This PR lacks support for: mqtt, sns, slack, and prometheus-alertmanager. It also leaves more in-depth client refactoring to followups.

Most changes to tests are to facilitate the new `factory_test` that verifies a provided custom dialer blocking requests works as expected. Various issues with the "valid" testing configs for integrations were causing them to fail for other reasons before they hit the dialer.


http/client.go
```diff
@@ -18,23 +20,53 @@ import (
...
+var (
+	defaultClientConfiguration = func() clientConfiguration {
+		return clientConfiguration{
+			userAgent: "Grafana",
+			dialer: (net.Dialer{
+				Timeout: 30 * time.Second,
+			}),
+		}
+	}
+)
```
rwwiv: We can just pull this into the `NewClient` constructor now.

http/client.go
```diff
@@ -18,23 +20,53 @@ import (
...
+
+func WithUserAgent(userAgent string) ClientOption {
+	return func(c *clientConfiguration) {
+		c.userAgent = userAgent
+	}
+}
+
+func WithDialer(dialer net.Dialer) ClientOption {
+	return func(c *clientConfiguration) {
+		c.dialer = dialer
```
rwwiv: Any reason we're using a mutator pattern here instead of an options struct?

JacobsonMT: Mostly because upstream uses a mutator pattern in its equivalent implementation. Figured there was some small benefit to keeping the implementations similar.

Other than that:
- It makes the usage pattern fairly clear. For example, that the configuration is optional, I have a slight preference towards this over passing in `nil` or exporting a `DefaultClientConfiguration` struct.
- Allows us to more easily provide selective modifications to the client per integration when necessary. For example, letting just `Jira` perform `GET` requests, or if we move `TLS`/`HMAC` config to `ClientOption` it can allows us to gradually implement support for those in other receivers without needing a `ClientConfiguration` "clone"/"apply option" flow.

These are not super important though, if you have a strong preference.

rwwiv: To your points:
- You can abuse variadic parameters to avoid needing to pass in a `nil` or default option (e.g. `... Option`) 
- I'm not sure I follow here, either method lets us do this no? A new field on the option struct would accomplish the same thing that another mutator would.


Here's an example of how this would look:

```go
type Client struct {
    log logging.Logger
    cfg clientConfiguration
}

type clientConfiguration struct {
    userAgent string
    dialer    *net.Dialer
}

type Options struct {
    UserAgent string
    Dialer    *net.Dialer
}

func NewClient(log logging.Logger, opts ...Options) *Client {
    // Default configuration
    cfg := clientConfiguration{
        userAgent: "Grafana",
        dialer: &net.Dialer{
            Timeout: 30 * time.Second,
        },
    }

    // Apply options if provided
    if len(opts) > 0 {
        opt := opts[0]
        if opt.UserAgent != "" {
            cfg.userAgent = opt.UserAgent
        }
        if opt.Dialer != nil {
            cfg.dialer = opt.Dialer
        }
    }

    return &Client{
        log: log,
        cfg: cfg,
    }
}

func (c *Client) Dialer() *net.Dialer {
    return c.cfg.dialer
}
```

JacobsonMT: Thanks for the example, that’s helpful.

I see what you mean, and you're right that both patterns support selective configuration in the end. Personally, I tend to dislike using variadic parameters to simulate optionality, since it can feel like an abuse of the language and sometimes obscure usage expectations for callers.

That said, I don't think there's a huge technical advantage either way. It's mostly a stylistic choice, and I’m happy to go with the options struct if you have a strong preference. Just let me know.

http/client.go
```diff
@@ -18,23 +20,47 @@ import (
...
-func NewClient(log logging.Logger, cfg ClientConfiguration) *Client {
+func NewClient(log logging.Logger, opts ...ClientOption) *Client {
+	cfg := clientConfiguration{
+		userAgent: "Grafana",
+		dialer: (net.Dialer{
+			Timeout: 30 * time.Second,
+		}),
+	}
+	for _, opt := range opts {
+		opt(&cfg)
```
yuri-tceretian: I think we should be a little more defensive and check for `opt` to be nil here

http/client.go
```diff
@@ -18,23 +20,47 @@ import (
...
 	return &Client{
-		log:   log,
-		agent: cfg.UserAgent,
+		log: log,
+		cfg: cfg,
+	}
+}
+
+func (ns *Client) Dialer() *net.Dialer {
+	return &ns.cfg.dialer
```
yuri-tceretian: Can we return copy of the dialer instead? Returning pointer allows client to be implicitly mutated.

http/client.go
```diff
@@ -18,23 +20,47 @@ import (
...
+	log logging.Logger
+	cfg clientConfiguration
 }
 
-func NewClient(log logging.Logger, cfg ClientConfiguration) *Client {
+func NewClient(log logging.Logger, opts ...ClientOption) *Client {
+	cfg := clientConfiguration{
+		userAgent: "Grafana",
+		dialer: (net.Dialer{
+			Timeout: 30 * time.Second,
```
yuri-tceretian: parenthesis seem to be unnecessary. Where does this 30s come from?

http/tls.go
```diff
@@ -1,31 +1,34 @@
...
+	if dialContextfunc == nil {
+		dialContextfunc = (&net.Dialer{
+			Timeout: 30 * time.Second,
+		}).DialContext
 	}
 
-	return nc(tlsConfig)
+	return &http.Client{
+		Timeout: time.Second * 30,
+		Transport: &http.Transport{
```
yuri-tceretian, line 27: Perhaps we should set IdleConnTimeout here?

http/tls.go
```diff
@@ -1,31 +1,34 @@
...
-		return nc(&tls.Config{Renegotiation: tls.RenegotiateFreelyAsClient})
+	if dialContextfunc == nil {
+		dialContextfunc = (&net.Dialer{
+			Timeout: 30 * time.Second,
+		}).DialContext
 	}
 
-	return nc(tlsConfig)
+	return &http.Client{
+		Timeout: time.Second * 30,
```
yuri-tceretian, line 26: This way we basically cap all downstream timeouts handshake, sending request, receiving headers and body  etc, including dialer. I think this could lead to some hidden problems. Why do we set it to 30s? And if we do need it, probably we should control other timeouts too and do some coordination.

yuri-tceretian:
I am not sure I understand where the hard-coded timeouts came from. Also, I am worried about coordination of hard-coded and configured timeouts. Other than that, LGTM

http/client.go
JacobsonMT: Not sure, possibly from the `DefaultTransport` in stlib? Mostly just trying to keep unrelated functional changes to a minimum, so kept the `30s` default. Our client timeout is also `30s` which should cover ensuring the dial timeout remains bounded even when a custom dialer is provided.

http/tls.go
JacobsonMT: Similar to above, I agree our client settings need improvement but I don't want to change them in this PR.

http/tls.go
JacobsonMT: Similar to above, I agree our client settings need improvement but I don't want to change them in this PR.

http/client.go
yuri-tceretian: ok, let's annotate those constants with comments and mention the reason why they are there

yuri-tceretian (approved):
LGTM

http/client.go
```diff
@@ -18,23 +20,49 @@ import (
...
-		log:   log,
-		agent: cfg.UserAgent,
+		log: log,
+		cfg: cfg,
+	}
+}
+
+func (ns *Client) Dialer() net.Dialer {
+	return ns.cfg.dialer
+}
```
santihernandezc: Is there a plan to use this method in other repos? I couldn't find any calls here.

notify/factory_test.go
```diff
@@ -72,6 +76,56 @@ func TestBuildReceiverIntegrations(t *testing.T) {
 		t.Run("should call notify wrapper for each config", func(t *testing.T) {
 			require.Equal(t, qty, wrapped)
 		})
+		t.Run("should use custom dial context", func(t *testing.T) {
+			customDialError := fmt.Errorf("custom dial function error")
+			clientOpts := []http.ClientOption{
+				http.WithUserAgent("Grafana-test"),
+				http.WithDialer(net.Dialer{
+					// Also override the Resolver so that configuration that doesn't use real hostnamesalso return
+					// "custom dial function error" instead of "no such dns".
```
santihernandezc: Super nit, there's a space missing.
```suggestion
					// Also override the Resolver so that configuration that doesn't use real hostnames
					// also return "custom dial function error" instead of "no such dns".
```

santihernandezc (approved):
LGTM!
I agree with Yuri's comments about timeouts, but I understand this PR might not be the best place to fix it. I created an [issue to make timeouts configurable](https://github.com/grafana/alerting/issues/324).

http/client.go
JacobsonMT: It's used in the followup PR. Accidentally left it in this one.

http/client.go
JacobsonMT: > Our client timeout is also 30s which should cover ensuring the dial timeout remains bounded even when a custom dialer is provided.

On second thought, I'm no longer sure this makes sense. So, I've added a fallback to ensure the dial timeout is configured if given a custom dialer without one set. Maybe it's not necessary, but it won't hurt.

Added some comments as well.

PR #319, opened 2025-05-02
Support SingleEmail=false in the email sender

alexander-akhmetov:
Current implementation always sends one email for all recipients, and this PR adds support for `SingleEmail` configuration parameter. 

Related to https://github.com/grafana/support-escalations/issues/16087

### DIFF
--- a/receivers/email_sender.go
+++ b/receivers/email_sender.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/tls"
 	"embed"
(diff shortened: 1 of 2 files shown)


santihernandezc (approved):
Nice! :rocket: 
I tested it locally and it works as expected. With `SingleEmail`, it groups all emails into one with multiple recipients. Without `SingleEmail`, it sends one email per recipient.

PR #322, opened 2025-05-05
Remove unused if in `default_template.go`

renepupil:


renepupil:
I strongly suggest splitting the template into multiple lines, just like regular code:

```
{{- define "default.message" -}}
    {{- if gt (len .Alerts.Firing) 0 -}}
        **Firing**
        {{- template "__text_alert_list" .Alerts.Firing -}}
    {{- end -}}
    {{- if gt (len .Alerts.Resolved) 0 -}}
        **Resolved**
        {{- template "__text_alert_list" .Alerts.Resolved -}}
    {{- end -}}
{{- end -}}
```

renepupil:
MEMO: Same also in `{{- define "teams.default.message" -}}` block.

PR #325, opened 2025-05-08
Support merging two Alertmanager configurations

yuri-tceretian:
This pull request adds the `Merge` function in the `definition` package to combine two Alertmanager configurations into a unified configuration. The function merges receivers, time intervals, routes, and inhibit rules from a primary and secondary configuration while ensuring compatibility. 
It lets user specify DedupSuffix for renaming conflicting resources and SubtreeMatchers to isolate routes and inhibition rules from the secondary configuration. The Merge function deduplicates resources by appending suffixes (e.g., _secondary or _01) and tracks renamed resources in the MergeResult. 
It checks the target routing tree for potentially conflicting routes and returns errors if merging source routing tree would break the  target. 

Related to https://github.com/grafana/alerting-squad/issues/1113


definition/alertmanager.go
```diff
@@ -118,6 +118,30 @@ func AsGrafanaRoute(r *config.Route) *Route {
...
+
+	for ln, lv := range r.MatchRE {
+		matcher, err := labels.NewMatcher(labels.MatchRegexp, ln, lv.String())
+		if err != nil {
+			return nil, err
+		}
+		matchers = append(matchers, matcher)
+	}
+
+	matchers = append(matchers, append(r.Matchers, r.ObjectMatchers...)...)
```
yuri-tceretian, line 141: this code is taken from the upstream https://github.com/grafana/prometheus-alertmanager/blob/main/dispatch/route.go#L94-L117

definition/alertmanager.go
```diff
@@ -257,21 +281,8 @@ func (c *PostableApiAlertingConfig) UnmarshalYAML(unmarshal func(interface{}) er
 func (c *PostableApiAlertingConfig) Validate() error {
 	receivers := make(map[string]struct{}, len(c.Receivers))
 
-	var hasGrafReceivers, hasAMReceivers bool
```
yuri-tceretian, line 260: This should be safe now because in Grafana we have full control over the integrations and it is not possible to define upstream integration in the current APIs

definition/merge.go
```diff
@@ -0,0 +1,337 @@
...
+//	  SubtreeMatchers: config.Matchers{
+//	    {Name: "source", Value: "external", Type: labels.MatchEqual},
+//	  },
+//	})
+//
+// If the merge fails with ErrSubtreeMatchersConflict, it indicates that the SubtreeMatchers
+// would conflict with existing routes, potentially disrupting alert notifications. In this
+// case, you should choose different SubtreeMatchers that don't overlap with existing route
+// matchers.
+func Merge(a, b PostableApiAlertingConfig, opts MergeOpts) (MergeResult, error) {
```
stevesg, line 133: What would be the downsides of creating a new root, and putting both trees under it with continue=true? Is it primarily because it would be hard to define the default receiver?

stevesg (approved):
This looks great, can't see anything wrong with the logic.

definition/merge.go
yuri-tceretian: The benefits are not clear but in that case we will have to determine the new fallback receiver for the root as well as we the "root matcher" for the Grafana's sub-tree.

PR #327, opened 2025-05-12
Refactor templates package

yuri-tceretian:
This pull request refactors package `templates` and makes it more isolate exposing less implementation details outside, making private as many functions as possible.
The goal of this pull request is to prepare the package to provide a single way to obtain Grafana or Prometheus templates, which will be implemented later.

Key changes:
- introduced template `Kind`, which can be either Grafana or Prometheus. 
- refactored fromContent function to build kind-agnostic templates.
- moved functions to process templates from notify package and made them public. 
- updated function TemplateFromContent to accept TemplateDefinition instead of list of template content

PR #328, opened 2025-05-12
Template factory

yuri-tceretian:
This PR is continuation of refactoring https://github.com/grafana/alerting/pull/327. 
- It replaces function TemplateFromTemplateDefinitions by a Factory that provides the templates depending on kind. 
- A new field Kind is added to TemplateDefinition, which lets us combine definitions of different kinds.

Usages are updated to use factory, which effectively allows creating Prometheus or Grafana template depending on context. 

Related to https://github.com/grafana/alerting-squad/issues/1113

### DIFF
--- a/templates/factory.go
+++ b/templates/factory.go
@@ -38,7 +38,6 @@ func (tp *Factory) NewTemplate(kind Kind, options ...template.Option) (*Template
 }
 
 // WithTemplate creates a new factory that has the provided TemplateDefinition. If definition with the same name already exists for this kind, it is replaced.
-// If TemplateDefinition.Kind is not known, GrafanaKind automatically assumed.
 func (tp *Factory) WithTemplate(def TemplateDefinition) (*Factory, error) {
 	if err := ValidateKind(def.Kind); err != nil {
 		return nil, err
--- a/templates/factory.go
+++ b/templates/factory.go
@@ -1,0 +1,104 @@
+package templates
+
+import (
+	"fmt"
+	"net/url"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/prometheus/alertmanager/template"
+)
+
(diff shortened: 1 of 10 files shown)


templates/template_data.go
```diff
@@ -77,6 +77,22 @@ type TemplateDefinition struct {
 	Kind Kind
 }
 
+func (t TemplateDefinition) Validate() error {
```
yuri-tceretian, line 95: It will be needed in grafana\gafana to check template validity in API

templates/factory.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	if tp.externalURL != nil {
+		t.ExternalURL = new(url.URL)
+		*t.ExternalURL = *tp.externalURL
+	}
+	return t, nil
+}
+
+// WithTemplate creates a new factory that has the provided TemplateDefinition. If definition with the same name already exists for this kind, it is replaced.
+// If TemplateDefinition.Kind is not known, GrafanaKind automatically assumed.
```
alexander-akhmetov: The methods actually fail in `ValidateKind` if the kind is unknown. As far as I see, the automatic assumption only happens in `TestTemplate`. Could we update the comments in factory.go?

alexander-akhmetov (approved):
LGTM

yuri-tceretian:
@JacobsonMT brought up an interesting side-effect that caching would cause: apparently, not all gomplate's functions are immutable. There is [one](https://github.com/grafana/alerting/blob/cad0d384d430bcabd30d6fb665ad78b6750d5428/templates/gomplate/template.go#L45-L65) that has a side-effect. So, for now we decided to rebuild the template for testing receivers and template definitions, and then address it internally in the factory that would return cloned template

JacobsonMT (approved):
LGTM, though probably worth giving it a more thorough manual test in the grafana-side PR.

PR #329, opened 2025-05-16
Refactor Grafana Alertmanager

yuri-tceretian:
This pull request introduces significant refactoring and enhancements to the `GrafanaAlertmanager` implementation in the `notify` package. The changes focus on improving configuration management, adding new dependencies, and simplifying the construction of `GrafanaAlertmanager` instances. Key updates include replacing the `GrafanaAlertmanagerConfig` struct with `GrafanaAlertmanagerOpts`, introducing new fields and validation logic, and refactoring the configuration application process to improve maintainability and extensibility.

### Refactoring and Configuration Updates:
* Replaced `GrafanaAlertmanagerConfig` with `GrafanaAlertmanagerOpts`, which includes additional fields such as `EmailSender`, `ImageProvider`, `Decrtypter`, `LoggerFactory`, `Version`, `TenantKey`, and `TenantID`. Most of those fields are needed to build notifiers. Added validation logic to ensure all required fields are present.
* Refactored `ApplyConfig` to use the new `NotificationsConfiguration` struct, which consolidates configuration fields like `RoutingTree`, `InhibitRules`, and `Receivers`. Removed the `BuildReceiverIntegrationsFunc` method in favor of a new `buildReceiverIntegrations` method which is the carbon copy of the method in Grafana repository.

### Dependency Injection Enhancements:
* Added support for injecting dependencies like `EmailSender`, `ImageProvider`, and `LoggerFactory` into `GrafanaAlertmanager` to improve testability and flexibility.

### Constructor Simplification:
* Simplified the `NewGrafanaAlertmanager` constructor by replacing multiple parameters with the new `GrafanaAlertmanagerOpts` struct. This change reduces the complexity of creating `GrafanaAlertmanager` instances.

### DIFF
--- a/notify/grafana_alertmanager.go
+++ b/notify/grafana_alertmanager.go
@@ -112,7 +112,7 @@ type GrafanaAlertmanager struct {
 
 	emailSender   receivers.EmailSender
 	imageProvider images.Provider
-	decrtypter    GetDecryptedValueFn
+	decrypter     GetDecryptedValueFn
 	loggerFactory logging.LoggerFactory
 
 	version string
@@ -211,7 +211,7 @@ func (c *GrafanaAlertmanagerOpts) Validate() error {
 	}
 
 	if c.Decrtypter == nil {
-		return errors.New("decrtypter must be present")
+		return errors.New("decrypter must be present")
 	}
 
 	if c.LoggerFactory == nil {
@@ -255,7 +255,7 @@ func NewGrafanaAlertmanager(opts GrafanaAlertmanagerOpts) (*GrafanaAlertmanager,
 		tenantID:          opts.TenantID,
 		externalURL:       opts.ExternalURL,
 		loggerFactory:     opts.LoggerFactory,
-		decrtypter:        opts.Decrtypter,
+		decrypter:         opts.Decrtypter,
 		version:           opts.Version,
 		emailSender:       opts.EmailSender,
 		imageProvider:     opts.ImageProvider,
@@ -976,7 +976,7 @@ func (am *GrafanaAlertmanager) tenantString() string {
 }
 
 func (am *GrafanaAlertmanager) buildReceiverIntegrations(receiver *APIReceiver, tmpl *templates.Template) ([]*Integration, error) {
-	receiverCfg, err := BuildReceiverConfiguration(context.Background(), receiver, DecodeSecretsFromBase64, am.decrtypter)
+	receiverCfg, err := BuildReceiverConfiguration(context.Background(), receiver, DecodeSecretsFromBase64, am.decrypter)
 	if err != nil {
 		return nil, err
 	}
--- a/notify/grafana_alertmanager.go
+++ b/notify/grafana_alertmanager.go
@@ -70,16 +70,12 @@ type ClusterPeer interface {
 }
 
 type GrafanaAlertmanager struct {
-	logger  log.Logger
-	Metrics *GrafanaAlertmanagerMetrics
+	opts   GrafanaAlertmanagerOpts
+	logger log.Logger
 
-	tenantID int64
-
-	marker      types.Marker
-	alerts      *mem.Alerts
-	route       *dispatch.Route
-	peer        ClusterPeer
-	peerTimeout time.Duration
+	marker types.Marker
+	alerts *mem.Alerts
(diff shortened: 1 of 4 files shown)


notify/grafana_alertmanager.go
```diff
@@ -102,12 +105,17 @@ type GrafanaAlertmanager struct {
 	config          []byte
 	receivers       []*nfstatus.Receiver
 
-	// buildReceiverIntegrationsFunc builds the integrations for a receiver based on its APIReceiver configuration and the current parsed template.
-	buildReceiverIntegrationsFunc func(next *APIReceiver, tmpl *templates.Template) ([]*Integration, error)
```
titolins, line 106: Tks! Never quite understood why so many wraps 😅

notify/grafana_alertmanager.go
```diff
@@ -102,12 +105,17 @@ type GrafanaAlertmanager struct {
...
-	externalURL                   string
+	externalURL string
 
 	// templates contains the template name -> template contents for each user-defined template.
 	templates []templates.TemplateDefinition
+
+	emailSender   receivers.EmailSender
+	imageProvider images.Provider
+	decrtypter    GetDecryptedValueFn
```
titolins: typo:
```suggestion
	decrypter    GetDecryptedValueFn
```

notify/grafana_alertmanager.go
```diff
@@ -184,39 +202,75 @@ func (c *GrafanaAlertmanagerConfig) Validate() error {
...
+		return errors.New("metrics must be present")
+	}
+
 	return nil
 }
 
 // NewGrafanaAlertmanager creates a new Grafana-specific Alertmanager.
-func NewGrafanaAlertmanager(tenantKey string, tenantID int64, config *GrafanaAlertmanagerConfig, peer ClusterPeer, logger log.Logger, m *GrafanaAlertmanagerMetrics) (*GrafanaAlertmanager, error) {
-	// TODO: Remove the context.
+func NewGrafanaAlertmanager(opts GrafanaAlertmanagerOpts) (*GrafanaAlertmanager, error) {
```
titolins, line 228: One nit about the naming is that opts usually refers to the opts functional pattern, which might be a good fit here actually. e.g.;
```suggestion
func NewGrafanaAlertmanager(opts ...func(*GrafanaAlertmanagerOpts)) (*GrafanaAlertmanager, error) {
    amOpts := GrafanaAlertmanagerOpts{
        // set some default values
    }
    for _, o := range opts {
        o(amOpts)
    }
```

`GrafanaAlertmanagerOpts` could be made private also, so we abstract the main configuration from the interface and possibly simplify things a bit more.. not sure it's worth the effort at all, just an idea 👍

notify/grafana_alertmanager.go
```diff
@@ -184,39 +202,75 @@ func (c *GrafanaAlertmanagerConfig) Validate() error {
...
-		externalURL:       config.ExternalURL,
-	}
-
-	if err := config.Validate(); err != nil {
-		return nil, err
+		logger:            log.With(opts.Logger, "component", "alertmanager", opts.TenantKey, opts.TenantID),
+		marker:            types.NewMarker(opts.Metrics.Registerer),
+		stageMetrics:      notify.NewMetrics(opts.Metrics.Registerer, featurecontrol.NoopFlags{}),
+		dispatcherMetrics: dispatch.NewDispatcherMetrics(false, opts.Metrics.Registerer),
+		peer:              opts.Peer,
```
titolins: It might make sense to just add opts to the alertmanager struct instead of having to remap the fields 1:1 - wdyt?

titolins (approved):
LGTM; just a couple of nits

notify/grafana_alertmanager.go
yuri-tceretian: I called it opts because there was too many "settings" for the same struct. I hesitated to use that pattern because it would bring some unnecessary flexibility and would assume implementation of the factory functions for each field of the options struct to complete it. 
On the second though, it could make it clearer the setting of options.

yuri-tceretian: I decided to not go with this as this pattern assumes that fields have sensible defaults. In our case they're not. So, the usage will become pretty verbose.

PR #331, opened 2025-05-20
BuildReceiverIntegrations to use single email sender

yuri-tceretian:
This PR refactors function BuildReceiverIntegrations to accept only single EmailSender instead of a factory function.
This simplifies the function interfaces and its usages.

### DIFF
--- a/receivers/email_sender.go
+++ b/receivers/email_sender.go
@@ -44,7 +44,7 @@ type defaultEmailSender struct {
 	dialFn func(*defaultEmailSender) (gomail.SendCloser, error)
 }
 
-// NewEmailSender takes a configuration and returns a new EmailSender factory function.
+// NewEmailSender takes a configuration and returns a new EmailSender.
 func NewEmailSender(cfg EmailSenderConfig) (EmailSender, error) {
 	tmpl, err := template.New("templates").
 		Funcs(template.FuncMap{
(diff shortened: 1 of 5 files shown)


receivers/email_sender.go
```diff
@@ -44,26 +44,24 @@ type defaultEmailSender struct {
...
-		}
-		return &defaultEmailSender{
-			cfg:  cfg,
-			tmpl: tmpl,
-			dialFn: func(s *defaultEmailSender) (gomail.SendCloser, error) {
-				return s.dial()
-			},
-		}, nil
+// NewEmailSender takes a configuration and returns a new EmailSender factory function.
+func NewEmailSender(cfg EmailSenderConfig) (EmailSender, error) {
```
yuri-tceretian, line 48: This sender is thread safe and there is no need to have many instances of it and the template.

alexander-akhmetov (approved):

receivers/email_sender.go
```diff
@@ -44,26 +44,24 @@ type defaultEmailSender struct {
...
-			return nil, err
-		}
-		return &defaultEmailSender{
-			cfg:  cfg,
-			tmpl: tmpl,
-			dialFn: func(s *defaultEmailSender) (gomail.SendCloser, error) {
-				return s.dial()
-			},
-		}, nil
+// NewEmailSender takes a configuration and returns a new EmailSender factory function.
```
santihernandezc: I'd remove the "factory function" at the end.

santihernandezc (approved):
LGTM! 
Just a small nit.

PR #337, opened 2025-06-02
Copy Mimir templates

yuri-tceretian:
This PR copies additional functions from Mimir into alerting.
https://github.com/grafana/mimir/blob/9757b8fed9a2482dee5bdf01367ef868601ed263/pkg/alertmanager/alertmanager_template.go

 It adds those functions to both kinds as it's done in Mimir now 
https://github.com/grafana/mimir/blob/a14e502853e077d57d8541a729b65f06386d9352/pkg/alertmanager/alertmanager.go#L643-L649

### DIFF
--- a/templates/funcs.go
+++ b/templates/funcs.go
@@ -19,7 +19,7 @@ func defaultTemplatesPerKind(kind Kind) []string {
 	}
 }
 
-func defaultOptionsPerKind(kind Kind) []template.Option {
+func defaultOptionsPerKind(kind Kind, orgID string) []template.Option {
 	switch kind {
 	case GrafanaKind:
 		return []template.Option{
--- a/templates/funcs.go
+++ b/templates/funcs.go
@@ -26,6 +26,7 @@ func defaultOptionsPerKind(kind Kind, orgID string) []template.Option {
 	case GrafanaKind:
 		return []template.Option{
 			addFuncs,
+			mimir.WithCustomFunctions(orgID),
 		}
 	case MimirKind:
 		return []template.Option{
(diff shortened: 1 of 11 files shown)


templates/template_data.go
```diff
@@ -94,7 +93,7 @@ func (t TemplateDefinition) Validate() error {
 	}
 	// Validate template contents. We try to stick as close to what will actually happen when the templates are parsed
 	// by the alertmanager as possible.
-	tmpl, err := template.New(defaultOptionsPerKind(t.Kind)...)
+	tmpl, err := template.New(defaultOptionsPerKind(t.Kind, "grafana")...)
```
titolins, line 96: just a nit, but does it make sense to create a variable for this?

titolins (approved):

templates/template_data.go
yuri-tceretian: it's just a placeholder. It can be any string. I did not want to make it anything more by declaring a shared constant

templates/funcs.go
```diff
@@ -19,11 +21,16 @@ func defaultTemplatesPerKind(kind Kind) []string {
 	}
 }
 
-func defaultOptionsPerKind(kind Kind) []template.Option {
+func defaultOptionsPerKind(kind Kind, orgID string) []template.Option {
```
alexander-akhmetov, line 24: should it maybe be userID/tenantID?

yuri-tceretian: Good question! I am actually on the fence about naming here. We already use term orgID in other places of this module. Introducing something different would be confusing for code reader. From another point of view, the function we introduced uses `tenantID` term, and changing it would be a breaking change. 
I decided to keep the term in code and in function as is.

alexander-akhmetov (approved):

PR #338, opened 2025-06-02
Add OAuth2 Support for HTTP Client Configuration in Webhook Receivers

JacobsonMT:
This PR introduces OAuth2 `client_credentials` grant type support for HTTP client configurations in webhook receivers.

### Current vs Future Work

This PR introduces `HTTPClientConfig` with a single field  `OAuth2` containing the new OAuth2 configuration options. This setup intentionally mimics upstream alertmanager's common `http_config` as the intention is to bring this change to all relevant integrations and eventually include other common http client configuration options in a standardized way between notifier types.

### All added configuration fields:

```yaml
http_config:
  oauth2:
    client_id: "test-client-id"
    client_secret: "test-client-secret"
    token_url: "https://localhost/auth/token"
    scopes:
      - "scope1"
      - "scope2"
    endpoint_params:
      param1: "value1"
      param2: "value2"
    tls_config:
      insecureSkipVerify: false
      caCertificate: "-----BEGIN CERTIFICATE..."
      clientCertificate: "-----BEGIN CERTIFICATE..."
      clientKey: "-----BEGIN EC PRIVATE KEY..."
    proxy_config:
      proxy_url: "http://localproxy:8080"
      no_proxy: "localhost,1.2.3.4"
      proxy_from_environment: false
      proxy_connect_header:
        X-Proxy-Header: "proxy-value"
```

### Notable differences from upstream
- `oauth2`: Does not add support for:
  -  `client_secret_file`
  -  `client_secret_ref`
- `oauth2.tls_config`:
  - Does not add support for:
    - `ca_file`
    - `cert_file`
    - `key_file`
    - `ca_ref`
    - `cert_ref`
    - `key_ref`
    - `server_name`
    - `min_version`
    - `max_version`
  - Uses the existing `TLSConfig` struct. This means out json fields does not match. Ex: `cert` vs `clientCertificate`. I'm on the fence here on whether it's worth defining a new struct to keep the field namings the same, thoughts?
- `oauth2.tls_config`:
  - `proxy_connect_header` values are not secured in Grafana as secure map values are not supported. We can consider breaking similar with upstream and adding a field specifically for `Proxy-Authorization`, thoughts?
  - Proxying is enabled (compared to Mimir AM) as we apply to firewall dialer to the oauth2 client.

### Implementation Notes
- `OAuth2` config validation is done during client creation. This is for two reasons:
  - The `tokenSource` needs to be stored outside `Notify` as the token is re-used until expiration.
  - Prefer failing invalid configuration early when creating the contact point, instead of silently failing to send notifications afterwards.
- `BuildReceiverIntegrations`: The above change means this function needs to return an error again (it was recently removed).
- It will make sense to move/extend various existing configs to the new common `http_config`, but this is not done yet. For example, the webhook notifer already has `tlsConfig` and `hmacConfig` and it would make sense to move these to `http_config`. This will require a migration though, as the config path will change and the current implementation of these options fail on Send instead of on Create.
- Since this is not the simplest feature to test/validate I added vairly comprehensive unit tests. Local testing can be done with your choice of local or cloud oauth2 provider (ex. Keycloak, Ory Hydra, Auth0, ...)

#### OAuth2 Integration:
* Added support for OAuth2 in `http/client.go`, including a new `oauth2.TokenSource` field in the `Client` struct and validation for OAuth2 configurations during client initialization. (`[[1]](diffhunk://#diff-db6dcd51441efbbef50fb4f0e52a2a24d0207a8cf997c2c2d18f1345e0b4b255R28-R39)`, `[[2]](diffhunk://#diff-db6dcd51441efbbef50fb4f0e52a2a24d0207a8cf997c2c2d18f1345e0b4b255L50-R71)`)
* Introduced the `WithHTTPClientConfig` option to pass HTTP client configurations, including OAuth2 settings. (`[http/client.goR88-R95](diffhunk://#diff-db6dcd51441efbbef50fb4f0e52a2a24d0207a8cf997c2c2d18f1345e0b4b255R88-R95)`)
* Enhanced the `SendWebhook` method to include OAuth2 token handling via an OAuth2 round-tripper. (`[http/client.goR149-R153](diffhunk://#diff-db6dcd51441efbbef50fb4f0e52a2a24d0207a8cf997c2c2d18f1345e0b4b255R149-R153)`)

#### Proxy Configuration:
* Added a new `ProxyConfig` struct in `http/config.go` to manage proxy settings, including support for proxy URLs, no-proxy addresses, and headers for CONNECT requests. (`[http/config.goR1-R128](diffhunk://#diff-9cd5b98b32603671d06faab6568b1be0d3f2790b8ee426270cb20cb91824997aR1-R128)`)
* Implemented validation logic for proxy configurations to ensure proper usage of proxy-related fields. (`[http/config.goR1-R128](diffhunk://#diff-9cd5b98b32603671d06faab6568b1be0d3f2790b8ee426270cb20cb91824997aR1-R128)`)

#### Test Coverage Enhancements:
* Updated tests in `http/client_test.go` to validate OAuth2 integration, including scenarios for valid and invalid configurations, token reuse, and proxy interactions. (`[[1]](diffhunk://#diff-0a04f5d21c1d2223648ce05c57de7f0e6b9be9be2da51184c5b441a3602845e6R5-R81)`, `[[2]](diffhunk://#diff-0a04f5d21c1d2223648ce05c57de7f0e6b9be9be2da51184c5b441a3602845e6R234-R531)`)
* Added a comprehensive test suite for OAuth2, covering various configurations such as scopes, endpoint parameters, and custom dialers. (`[http/client_test.goR234-R531](diffhunk://#diff-0a04f5d21c1d2223648ce05c57de7f0e6b9be9be2da51184c5b441a3602845e6R234-R531)`)

### DIFF
--- a/http/config.go
+++ b/http/config.go
@@ -1,0 +1,128 @@
+package http
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+
+	"golang.org/x/net/http/httpproxy"
+
+	"github.com/grafana/alerting/receivers"
+)
+
+var (
+	// ErrInvalidOAuth2Config is returned when the OAuth2 configuration is invalid.
+	ErrInvalidOAuth2Config = fmt.Errorf("invalid OAuth2 configuration")
+	// ErrInvalidProxyConfig is returned when the proxy configuration is invalid.
+	ErrInvalidProxyConfig = fmt.Errorf("invalid proxy configuration")
+)
+
+// HTTPClientConfig holds common configuration for notifier HTTP clients.
+type HTTPClientConfig struct {
+	// The OAuth2 client credentials used to fetch a token for the targets.
+	OAuth2 *OAuth2Config `yaml:"oauth2,omitempty" json:"oauth2,omitempty"`
+}
+
+// Decrypt decrypts sensitive fields in the HTTPClientConfig.
+func (c *HTTPClientConfig) Decrypt(decryptFn receivers.DecryptFunc) {
+	if c.OAuth2 != nil {
+		c.OAuth2.ClientSecret = decryptFn("http_config.oauth2.client_secret", c.OAuth2.ClientSecret)
+		if c.OAuth2.TLSConfig != nil {
+			c.OAuth2.TLSConfig.CACertificate = decryptFn("http_config.oauth2.tls_config.caCertificate", c.OAuth2.TLSConfig.CACertificate)
+			c.OAuth2.TLSConfig.ClientCertificate = decryptFn("http_config.oauth2.tls_config.clientCertificate", c.OAuth2.TLSConfig.ClientCertificate)
+			c.OAuth2.TLSConfig.ClientKey = decryptFn("http_config.oauth2.tls_config.clientKey", c.OAuth2.TLSConfig.ClientKey)
+		}
+	}
+}
+
+func ValidateHTTPClientConfig(cfg *HTTPClientConfig) error {
+	if cfg != nil {
+		if err := ValidateOAuth2Config(cfg.OAuth2); err != nil {
+			return fmt.Errorf("%w: %w", ErrInvalidOAuth2Config, err)
+		}
+	}
+	return nil
+}
+
+type ProxyConfig struct {
+	// ProxyURL is the HTTP proxy server to use to connect to the targets.
+	ProxyURL string `yaml:"proxy_url,omitempty" json:"proxy_url,omitempty"`
+	// NoProxy contains addresses that should not use a proxy.
+	NoProxy string `yaml:"no_proxy,omitempty" json:"no_proxy,omitempty"`
+	// ProxyFromEnvironment uses environment HTTP_PROXY, HTTPS_PROXY and NO_PROXY to determine proxies.
+	ProxyFromEnvironment bool `yaml:"proxy_from_environment,omitempty" json:"proxy_from_environment,omitempty"`
+	// ProxyConnectHeader optionally specifies headers to send to proxies during CONNECT requests.
+	ProxyConnectHeader map[string]string `yaml:"proxy_connect_header,omitempty" json:"proxy_connect_header,omitempty"`
+}
+
+// Proxy returns the Proxy URL for a request.
+func (cfg *ProxyConfig) Proxy() (fn func(*http.Request) (*url.URL, error), err error) {
+	if cfg == nil {
+		return nil, nil
+	}
+	if cfg.ProxyFromEnvironment {
+		proxyFn := httpproxy.FromEnvironment().ProxyFunc()
+		return func(req *http.Request) (*url.URL, error) {
+			return proxyFn(req.URL)
+		}, nil
+	}
+	if cfg.ProxyURL != "" {
+		proxyUrl, err := url.Parse(cfg.ProxyURL)
+		if err != nil {
+			return nil, fmt.Errorf("invalid proxy URL %q: %w", cfg.ProxyURL, err)
+		}
+		if cfg.NoProxy == "" {
+			return http.ProxyURL(proxyUrl), nil
+		}
+		proxy := &httpproxy.Config{
+			HTTPProxy:  proxyUrl.String(),
+			HTTPSProxy: proxyUrl.String(),
+			NoProxy:    cfg.NoProxy,
+		}
+		proxyFn := proxy.ProxyFunc()
+		return func(req *http.Request) (*url.URL, error) {
+			return proxyFn(req.URL)
+		}, nil
+	}
+	return nil, nil
+}
+
+func (cfg *ProxyConfig) GetProxyConnectHeader() http.Header {
+	if cfg == nil || len(cfg.ProxyConnectHeader) == 0 {
+		return nil
+	}
+	// Return a copy of the header to avoid modifying the original.
+	headerCopy := make(http.Header, len(cfg.ProxyConnectHeader))
+	for k, v := range cfg.ProxyConnectHeader {
+		headerCopy.Add(k, v)
+	}
+	return headerCopy
+}
+
+func ValidateProxyConfig(cfg *ProxyConfig) error {
+	if cfg == nil {
+		// If no proxy config is provided, we consider it valid.
+		return nil
+	}
+	if len(cfg.ProxyConnectHeader) > 0 && !cfg.ProxyFromEnvironment && cfg.ProxyURL == "" {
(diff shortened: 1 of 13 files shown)


http/config.go
```diff
@@ -0,0 +1,130 @@
...
+//
+//nolint:revive
+type HTTPClientConfig struct {
+	// The OAuth2 client credentials used to fetch a token for the targets.
+	OAuth2 *OAuth2Config `yaml:"oauth2,omitempty" json:"oauth2,omitempty"`
+}
+
+// Decrypt decrypts sensitive fields in the HTTPClientConfig.
+func (c *HTTPClientConfig) Decrypt(decryptFn receivers.DecryptFunc) {
+	if c.OAuth2 != nil {
```
yuri-tceretian: perhaps invert ifs?

http/config.go
```diff
@@ -0,0 +1,130 @@
...
+		return nil, nil
+	}
+	if cfg.ProxyFromEnvironment {
+		proxyFn := httpproxy.FromEnvironment().ProxyFunc()
+		return func(req *http.Request) (*url.URL, error) {
+			return proxyFn(req.URL)
+		}, nil
+	}
+	if cfg.ProxyURL != "" {
+		proxyURL, err := url.Parse(cfg.ProxyURL)
```
yuri-tceretian: WDYT about storing URL in ProxyConfig instead of string?

http/config.go
```diff
@@ -0,0 +1,130 @@
...
+		if err := ValidateOAuth2Config(cfg.OAuth2); err != nil {
+			return fmt.Errorf("%w: %w", ErrInvalidOAuth2Config, err)
+		}
+	}
+	return nil
+}
+
+type ProxyConfig struct {
+	// ProxyURL is the HTTP proxy server to use to connect to the targets.
+	ProxyURL string `yaml:"proxy_url,omitempty" json:"proxy_url,omitempty"`
```
yuri-tceretian: I wonder if it should be masked i.e. overridden .String() so we do not accidentally expose any urls

http/oauth2.go
```diff
@@ -0,0 +1,143 @@
...
+	var tlsConfig *tls.Config
+	if config.TLSConfig != nil {
+		var err error
+		tlsConfig, err = config.TLSConfig.ToCryptoTLSConfig()
+		if err != nil {
+			return nil, fmt.Errorf("%w: %w", ErrOAuth2TLSConfigInvalid, err)
+		}
+	}
+
+	proxyFn, err := config.ProxyConfig.Proxy()
```
yuri-tceretian: ProxyConfig could be nil.

http/oauth2.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create proxy function: %w", err)
```
yuri-tceretian: Should it be `ErrInvalidProxyConfig`?

http/oauth2.go
```diff
@@ -0,0 +1,143 @@
...
+	if clientConfig.userAgent != "" {
+		rt = commoncfg.NewUserAgentRoundTripper(clientConfig.userAgent, rt)
+	}
+
+	return credconfig.TokenSource(
+		context.WithValue(
+			context.Background(),
+			oauth2.HTTPClient,
+			&http.Client{
+				Transport: rt,
```
yuri-tceretian, line 132: perhaps set `Timeout` to 30s  here so we have a hard limit for the request?

yuri-tceretian (commented):
Still looking into it. Comments so far

notify/factory.go
```diff
@@ -47,89 +51,141 @@ func BuildReceiverIntegrations(
...
-		ci(i, cfg.Metadata, victorops.New(cfg.Settings, cfg.Metadata, tmpl, cli, img, logger, version))
+		ci(i, cfg.Metadata, func(cli *http.Client) notificationChannel {
+			return victorops.New(cfg.Settings, cfg.Metadata, tmpl, cli, img, logger, version)
+		})
 	}
 	for i, cfg := range receiver.WebhookConfigs {
-		ci(i, cfg.Metadata, webhook.New(cfg.Settings, cfg.Metadata, tmpl, cli, img, logger, orgID))
+		ci(i, cfg.Metadata, func(cli *http.Client) notificationChannel {
+			return webhook.New(cfg.Settings, cfg.Metadata, tmpl, cli, img, logger, orgID)
+		}, http.WithHTTPClientConfig(cfg.Settings.HTTPConfig))
```
yuri-tceretian, line 202: WDYT about providing client factory into the integration that will be able to create a client with individual settings? This way we can keep part that uses the config to the integration

http/config.go
```diff
@@ -0,0 +1,130 @@
...
+	// NoProxy contains addresses that should not use a proxy.
+	NoProxy string `yaml:"no_proxy,omitempty" json:"no_proxy,omitempty"`
+	// ProxyFromEnvironment uses environment HTTP_PROXY, HTTPS_PROXY and NO_PROXY to determine proxies.
+	ProxyFromEnvironment bool `yaml:"proxy_from_environment,omitempty" json:"proxy_from_environment,omitempty"`
+	// ProxyConnectHeader optionally specifies headers to send to proxies during CONNECT requests.
+	ProxyConnectHeader map[string]string `yaml:"proxy_connect_header,omitempty" json:"proxy_connect_header,omitempty"`
```
yuri-tceretian, line 65: WDYT about making all header's values secure?

yuri-tceretian (approved):
LGTM 👍🏻

http/config.go
JacobsonMT: Possibly, though it's unclear at a glance where we'd want to redact the information. I can tackle in a followup.

notify/factory.go
JacobsonMT: Agreed, `WithHTTPClientConfig` will very likely be removed in the follow-up PR that extends OAuth support to the other notifiers. Initially, I took this approach to avoid making broader changes to the factory before settling on the right abstraction. In retrospect, though, there wasn’t much value in delaying it, and handling it up front would have made the transition cleaner.

http/config.go
JacobsonMT: I'd like to, but this is tougher than it seems. We don't really have a way to secure arbitrary `map[string]string` fields in `grafana/grafana` backend or a way to display that kind of component in the FE right now. I'll see what I can do in a followup.

saurabhb-dev:
Hi Team,

Can you please tell the estimate for UI to be released for this feature? In the current Grafana OSS there does not exists an UI for using this contact point.

Thanks,
Saurabh

PR #343, opened 2025-06-04
Templates factory to return clones

yuri-tceretian:
Update CachedTemplateFactory to return cloned templates. 
The interface of templates allows mutation - it exposes Parse function that adds more templates to the current tree. 
There is some code in `gomplate` package that uses this method, and therefore, it's not safe to share the same instance across multiple integrations.

### DIFF
--- a/templates/factory.go
+++ b/templates/factory.go
@@ -134,5 +134,5 @@ func (cf *CachedFactory) GetTemplate(kind Kind) (*Template, error) {
 		return nil, err
 	}
 	cf.m[kind] = t
-	return t, nil
+	return t.Clone()
 }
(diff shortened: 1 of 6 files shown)


titolins (approved):

templates/factory.go
```diff
@@ -134,5 +134,5 @@ func (cf *CachedFactory) GetTemplate(kind Kind) (*Template, error) {
 		return nil, err
 	}
 	cf.m[kind] = t
-	return t, nil
+	return t.Clone()
```
JacobsonMT, line 137: Need to return clone on line 130 as well

JacobsonMT (changes requested):

JacobsonMT (approved):

PR #349, opened 2025-06-22
Add custom JSON marshaller for Alertmanager configurations

alexander-akhmetov:
This PR adds a custom JSON marshaler (`MarshalJSONWithSecrets`) that can serialize Alertmanager configurations with secrets in plain text rather than masked values. It uses `jsoniter` to create custom encoders for secret types from Prometheus and Alertmanager.

It is not used yet.

Prometheus Alertmanager uses a few types of secret structs:
- [prometheus/common Secret](https://github.com/prometheus/common/blob/75c3814dc66c571cc82cee2d3a6bf5f37ee73f1a/config/config.go#L28)
- [prometheus/alertmanager Secret](https://github.com/prometheus/alertmanager/blob/0ce3cfb962db3cbb1649d3e816a49a13c4036cd1/config/config.go#L56)
- [prometheus/alertmanager SecretURL](https://github.com/prometheus/alertmanager/blob/0ce3cfb962db3cbb1649d3e816a49a13c4036cd1/config/config.go#L142)

During marshaling, secrets in these structs are masked (this can be controlled only via a global variable), but sometimes we need to preserve them and this PR adds a custom marshaler for these situations.

Part of https://github.com/grafana/alerting-squad/issues/1152

### DIFF
--- a/definition/json.go
+++ b/definition/json.go
@@ -1,0 +1,85 @@
+package definition
+
+import (
+	"unsafe"
+
+	jsoniter "github.com/json-iterator/go"
+	"github.com/modern-go/reflect2"
+	amcfg "github.com/prometheus/alertmanager/config"
+	commoncfg "github.com/prometheus/common/config"
+)
+
+// secretEncoder encodes Secret to plain text JSON,
+// avoding the default masking behavior of the structure.
+type secretEncoder struct{}
+
+func (encoder *secretEncoder) Encode(ptr unsafe.Pointer, stream *jsoniter.Stream) {
+	stream.WriteString(getStr(ptr))
+}
+
+func (encoder *secretEncoder) IsEmpty(ptr unsafe.Pointer) bool {
+	return len(getStr(ptr)) == 0
+}
+
+func getStr(ptr unsafe.Pointer) string {
+	return *(*string)(ptr)
+}
+
+// secretEncoder encodes SecretURL to plain text JSON,
+// avoding the default masking behavior of the structure.
+type secretURLEncoder struct{}
+
+func (encoder *secretURLEncoder) Encode(ptr unsafe.Pointer, stream *jsoniter.Stream) {
+	url := getURL(ptr)
+	if url.URL != nil {
+		stream.WriteString(url.String())
+	} else {
+		stream.WriteNil()
+	}
+}
+
+func (encoder *secretURLEncoder) IsEmpty(ptr unsafe.Pointer) bool {
+	url := getURL(ptr)
+	return url.URL == nil
+}
+
+func getURL(ptr unsafe.Pointer) *amcfg.URL {
+	v := (*amcfg.SecretURL)(ptr)
+	url := amcfg.URL(*v)
+	return &url
+}
+
+func newPlainAPI() jsoniter.API {
+	api := jsoniter.ConfigCompatibleWithStandardLibrary
+
+	secretEnc := &secretEncoder{}
+	secretURLEnc := &secretURLEncoder{}
+
+	extension := jsoniter.EncoderExtension{
+		// Value types
+		reflect2.TypeOfPtr((*amcfg.Secret)(nil)).Elem():     secretEnc,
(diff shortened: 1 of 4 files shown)


definition/json.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+func (encoder *secretURLEncoder) isSecretURLEmpty(v amcfg.SecretURL) bool {
+	url := amcfg.URL(v)
+	return url.URL == nil || url.String() == ""
+}
+
+func newPlainAPI() jsoniter.API {
+	api := jsoniter.ConfigCompatibleWithStandardLibrary
```
alexander-akhmetov, line 53: https://github.com/json-iterator/go/blob/71ac16282d122fdd1e3a6d3e7f79b79b4cc3b50e/config.go#L54

definition/json.go
```diff
@@ -0,0 +1,120 @@
...
+	commoncfg "github.com/prometheus/common/config"
+)
+
+// secretEncoder encodes Secret to plain text JSON,
+// avoding the default masking behavior of the structure.
+type secretEncoder struct {
+	isPtr bool
+}
+
+func (encoder *secretEncoder) Encode(ptr unsafe.Pointer, stream *jsoniter.Stream) {
```
alexander-akhmetov, line 16: Looked at the examples here https://github.com/json-iterator/go/blob/master/extra/time_as_int64_codec.go

definition/json.go
```diff
@@ -0,0 +1,85 @@
...
+	"unsafe"
+
+	jsoniter "github.com/json-iterator/go"
+	"github.com/modern-go/reflect2"
+	amcfg "github.com/prometheus/alertmanager/config"
+	commoncfg "github.com/prometheus/common/config"
+)
+
+// secretEncoder encodes Secret to plain text JSON,
+// avoding the default masking behavior of the structure.
```
fayzal-g: Nit: 
```suggestion
// avoiding the default masking behavior of the structure.
```

definition/json.go
```diff
@@ -0,0 +1,85 @@
...
+func (encoder *secretEncoder) IsEmpty(ptr unsafe.Pointer) bool {
+	return len(getStr(ptr)) == 0
+}
+
+func getStr(ptr unsafe.Pointer) string {
+	return *(*string)(ptr)
+}
+
+// secretEncoder encodes SecretURL to plain text JSON,
+// avoding the default masking behavior of the structure.
```
fayzal-g: Nit: 
```suggestion
// avoiding the default masking behavior of the structure.
```

definition/json.go
```diff
@@ -0,0 +1,85 @@
...
+type secretURLEncoder struct{}
+
+func (encoder *secretURLEncoder) Encode(ptr unsafe.Pointer, stream *jsoniter.Stream) {
+	url := getURL(ptr)
+	if url.URL != nil {
+		stream.WriteString(url.String())
+	} else {
+		stream.WriteNil()
+	}
```
fayzal-g, line 38: Nit: Use the `IsEmpty` function below?
```suggestion
	if !encoder.IsEmpty(ptr) {
		stream.WriteString(url.String())
	} else {
		stream.WriteNil()
	}
```

definition/json.go
```diff
@@ -0,0 +1,85 @@
...
+	return api
+}
+
+var (
+	plainJSON = newPlainAPI()
+)
+
+// MarshalJSONWithSecrets marshals the given value to JSON with secrets in plain text.
+//
+// alertmanager's and prometeus' Secret and SecretURL types mask their values
```
fayzal-g: Nit:
```suggestion
// alertmanager's and prometheus' Secret and SecretURL types mask their values
```

fayzal-g (approved):
Sorry for the ultra nits, logic is sound though, lgtm

definition/json.go
alexander-akhmetov: It still needs the `url`, to write the string

PR #352, opened 2025-07-07
Notification history

vstpme:
This PR introduces a concept of "Notification Historian". A notification historian is similar to a state historian used by Alert State History. Instead of capturing state transitions, a notification historian captures the result of integration's `Notify` method call.

This is useful for the ongoing Notification History project and will allow to write notification logs to Loki. The actual implementation of Loki notification historian will be added in https://github.com/grafana/grafana/pull/107644.

### DIFF
--- a/notify/nfstatus/integration.go
+++ b/notify/nfstatus/integration.go
@@ -11,7 +11,7 @@ import (
 )
 
 type NotificationHistorian interface {
-	Record(ctx context.Context, alerts []*types.Alert, notificationErr error) <-chan error
+	Record(ctx context.Context, alerts []*types.Alert, retry bool, notificationErr error, duration time.Duration) <-chan error
 }
 
 // Integration wraps an upstream notify.Integration, adding the ability to
@@ -99,7 +99,7 @@ func (n *statusCaptureNotifier) Notify(ctx context.Context, alerts ...*types.Ale
 	duration := time.Since(start)
 
 	if n.notificationHistorian != nil {
-		n.notificationHistorian.Record(ctx, alerts, err)
+		n.notificationHistorian.Record(ctx, alerts, retry, err, duration)
 	}
 
 	n.mtx.Lock()
--- a/notify/nfstatus/integration.go
+++ b/notify/nfstatus/integration.go
@@ -10,6 +10,10 @@ import (
 	"github.com/prometheus/common/model"
 )
 
+type NotificationHistorian interface {
+	Record(ctx context.Context, alerts []*types.Alert, notificationErr error) error
+}
+
 // Integration wraps an upstream notify.Integration, adding the ability to
 // capture notification status.
 type Integration struct {
@@ -18,9 +22,9 @@ type Integration struct {
 }
 
 // NewIntegration returns a new integration.
-func NewIntegration(notifier notify.Notifier, rs notify.ResolvedSender, name string, idx int, receiverName string) *Integration {
+func NewIntegration(notifier notify.Notifier, rs notify.ResolvedSender, name string, idx int, receiverName string, notificationHistorian NotificationHistorian) *Integration {
 	// Wrap the provided Notifier with our own, which will capture notification attempt errors.
-	status := &statusCaptureNotifier{upstream: notifier}
+	status := &statusCaptureNotifier{upstream: notifier, notificationHistorian: notificationHistorian}
 
 	integration := notify.NewIntegration(status, rs, name, idx, receiverName)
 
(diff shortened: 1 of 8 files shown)


notify/nfstatus/integration.go
```diff
@@ -10,6 +10,10 @@ import (
 	"github.com/prometheus/common/model"
 )
 
+type NotificationHistorian interface {
+	Record(ctx context.Context, alerts []*types.Alert, notificationErr error) <-chan error
+}
+
```
vstpme: this is similar to [`state.Historian`](https://github.com/grafana/grafana/blob/24ebb27ebeb54e6ab69ab26b36872f6cca6a7d7b/pkg/services/ngalert/state/persist.go#L40-L46)

notify/nfstatus/integration_test.go
```diff
@@ -76,3 +78,33 @@ func TestIntegration(t *testing.T) {
...
+type mockNotificationHistorian struct {
+	mock.Mock
+}
+
+func (m *mockNotificationHistorian) Record(ctx context.Context, alerts []*types.Alert, notificationErr error) <-chan error {
+	args := m.Called(ctx, alerts, notificationErr)
+	return args.Get(0).(chan error)
+}
+
+func TestIntegrationWithNotificationHistorian(t *testing.T) {
```
vstpme, line 91: adding a simple test to check that `Record` is called on `Notify`

receivers/jira/jira_test.go
```diff
@@ -542,9 +542,8 @@ func TestNotify(t *testing.T) {
 								transition,
 							},
 						}), 200)
-					} else {
-						return cmd.Validation(nil, 200)
 					}
+					return cmd.Validation(nil, 200)
```
vstpme, line 546: fixing random `make lint` issues

notify/nfstatus/integration.go
```diff
@@ -93,6 +98,10 @@ func (n *statusCaptureNotifier) Notify(ctx context.Context, alerts ...*types.Ale
 	retry, err := n.upstream.Notify(ctx, alerts...)
 	duration := time.Since(start)
 
+	if n.notificationHistorian != nil {
+		n.notificationHistorian.Record(ctx, alerts, err)
+	}
```
vstpme: the most important change of this PR

notify/nfstatus/integration.go
```diff
@@ -93,6 +98,10 @@ func (n *statusCaptureNotifier) Notify(ctx context.Context, alerts ...*types.Ale
 	retry, err := n.upstream.Notify(ctx, alerts...)
 	duration := time.Since(start)
 
+	if n.notificationHistorian != nil {
+		n.notificationHistorian.Record(ctx, alerts, err)
```
yuri-tceretian: Let's add more information here: Attempt # and duration. Perhaps retry status would be useful too.

yuri-tceretian: Also, I would make it clear here and run in a go-routine. It's very important that this piece is executed quickly, or in a non-blocking way, because otherwise it would affect the state propagation and cause pesky "flapping alerts".

yuri-tceretian (approved):
LGTM.

notify/nfstatus/integration.go
vstpme: **Re: adding more info**
Good call! I added retry status and duration in https://github.com/grafana/alerting/pull/352/commits/b8746f001d5c6ca8db8be1fdcf1fb2c49158e5d2. Unfortunately the attempt number is not passed down to `Notify`, I guess retries are handled somewhere else. I suggest we ignore it for now.

**Re: go routine**
`NotificationHistorian.Record` is supposed be async, that's why the return type is `<- chan error`. I agree it makes more sense to move the goroutine invocation here though. The plan is to move the notification historian implementation to `grafana/alerting` before going to prod (so it can be reused by `grafana/mimir`). I'll make sure to address this when I do that.

yuri-tceretian: > return type is <- chan error
1. I do not think returning an error makes sense because that error is not actionable. I would recommend returning void.
2. By calling it in a go routine you let the client control the flow rather than relying on assumption of implementation details.

yuri-tceretian: re: retry count 

you're right :( that information is not available at this level. However, I think it's important to have this info in the log (it will let us distinguish between duplicate notifications and retries. I think we can update RetryStage to include this information into the context and then we can retrieve it down the road.

vstpme: valid points 👍 I'll address it when moving the notification historian implementation to `grafana/alerting`

notify/nfstatus/integration.go
```diff
@@ -93,6 +98,10 @@ func (n *statusCaptureNotifier) Notify(ctx context.Context, alerts ...*types.Ale
 	retry, err := n.upstream.Notify(ctx, alerts...)
 	duration := time.Since(start)
 
+	if n.notificationHistorian != nil {
+		n.notificationHistorian.Record(ctx, alerts, retry, err, duration)
```
fayzal-g, line 102: I would check the return value in a follow-up PR

fayzal-g (approved):

PR #353, opened 2025-07-07
PostableApiTemplate

yuri-tceretian:
This PR introduces a new type in `definitions` package: PostableApiTemplate. This is an API model of templates.TemplateDefinition type and will be used along with `PostableApiAlertingConfig`. 

Currently, we use `map[string]string` for types which lacks the necessary information such as kind. 

Also, it introduces merge function to merge two sets of templates, which will be used along with merge for PostableApiAlertingConfig

### DIFF
--- a/definition/alertmanager.go
+++ b/definition/alertmanager.go
@@ -628,8 +628,8 @@ func (t *PostableApiTemplate) Validate() error {
 		return fmt.Errorf("template kind is required")
 	}
 	k := strings.ToLower(string(t.Kind))
-	if k != "grafana" && k != "mimir" {
-		return fmt.Errorf("invalid template kind, must be either 'grafana' or 'mimir'")
+	if k != string(GrafanaTemplateKind) && k != string(MimirTemplateKind) {
+		return fmt.Errorf("invalid template kind, must be either '%s' or '%s'", GrafanaTemplateKind, MimirTemplateKind)
 	}
 	return nil
 }
--- a/definition/alertmanager.go
+++ b/definition/alertmanager.go
@@ -609,3 +609,32 @@ func (pgr *PostableGrafanaReceiver) DecryptSecureSettings(decryptFn func(payload
 	}
 	return decrypted, nil
 }
+
+// nolint:revive
+type PostableApiTemplate struct {
+	Name    string       `yaml:"name" json:"name"`
+	Content string       `yaml:"content" json:"content"`
+	Kind    TemplateKind `yaml:"kind" json:"kind"`
+}
+
+func (t *PostableApiTemplate) Validate() error {
+	if t.Name == "" {
+		return fmt.Errorf("template name is required")
+	}
+	if t.Content == "" {
+		return fmt.Errorf("template content is required")
+	}
+	if t.Kind == "" {
+		return fmt.Errorf("template kind is required")
+	}
+	k := strings.ToLower(string(t.Kind))
+	if k != "grafana" && k != "mimir" {
+		return fmt.Errorf("invalid template kind, must be either 'grafana' or 'mimir'")
+	}
+	return nil
+}
+
+type TemplateKind string
+
+const GrafanaTemplateKind TemplateKind = "grafana"
+const MimirTemplateKind TemplateKind = "mimir"
--- a/definition/merge.go
+++ b/definition/merge.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"slices"
 
 	"github.com/prometheus/alertmanager/config"
 	"github.com/prometheus/alertmanager/dispatch"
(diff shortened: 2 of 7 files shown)


definition/alertmanager.go
```diff
@@ -609,3 +609,32 @@ func (pgr *PostableGrafanaReceiver) DecryptSecureSettings(decryptFn func(payload
...
+		return fmt.Errorf("template name is required")
+	}
+	if t.Content == "" {
+		return fmt.Errorf("template content is required")
+	}
+	if t.Kind == "" {
+		return fmt.Errorf("template kind is required")
+	}
+	k := strings.ToLower(string(t.Kind))
+	if k != "grafana" && k != "mimir" {
```
JacobsonMT: Should use the `TemplateKind` types here.

definition/merge.go
```diff
@@ -335,3 +336,26 @@ func getUniqueName(name string, suffix string, usedNames map[string]struct{}) st
...
 	return result
 }
+
+// MergeTemplates merges two slices of PostableApiTemplate, ensuring unique names and returns the merged slice and a map of renamed templates from slice `b`.
+func MergeTemplates(a, b []PostableApiTemplate, opts MergeOpts) ([]PostableApiTemplate, map[string]string) {
+	makeKey := func(name string, kind TemplateKind) string {
+		return fmt.Sprintf("%s|%s", name, kind)
+	}
+	usedNames := make(map[string]struct{}, len(a)+len(b))
+	result := slices.Grow(a, len(b))
```
JacobsonMT: I believe `result` and `a` will share a backing array, just to be safe we should probably create a new slice.

definition/merge.go
```diff
@@ -335,3 +336,26 @@ func getUniqueName(name string, suffix string, usedNames map[string]struct{}) st
 	}
 	return result
 }
+
+// MergeTemplates merges two slices of PostableApiTemplate, ensuring unique names and returns the merged slice and a map of renamed templates from slice `b`.
+func MergeTemplates(a, b []PostableApiTemplate, opts MergeOpts) ([]PostableApiTemplate, map[string]string) {
```
JacobsonMT: Note this makes the assumption that there are no duplicate names in the `b` input. This might be fine for how it's going to be used, but figured I'd point it out in case that's an issue.

definition/merge_test.go
```diff
@@ -446,6 +446,59 @@ func TestCheckIfMatchersUsed(t *testing.T) {
...
+			name:     "empty right slice should not rename anything",
+			left:     []PostableApiTemplate{{Name: "test", Kind: "grafana"}, {Name: "test", Kind: "grafana"}},
+			right:    nil,
+			expected: []PostableApiTemplate{{Name: "test", Kind: "grafana"}, {Name: "test", Kind: "grafana"}},
+			renamed:  map[string]string{},
+		},
+		{
+			name:     "empty left slice should only rename duplicates in right slice",
+			left:     nil,
+			right:    []PostableApiTemplate{{Name: "test", Kind: "grafana"}, {Name: "test", Kind: "grafana"}},
```
JacobsonMT: This is related to my comment above about duplicate names in `b`, do we want to be silently dropping one of the templates here?

notify/compat.go
```diff
@@ -26,3 +27,28 @@ func PostableAPIReceiverToAPIReceiver(r *definition.PostableApiReceiver) *APIRec
...
+	d := templates.TemplateDefinition{
+		Name:     t.Name,
+		Template: t.Content,
+		Kind:     kind,
+	}
+	return d
+}
+
+func PostableAPITemplatesToTemplateDefinitions(ts []definition.PostableApiTemplate) []templates.TemplateDefinition {
+	var defs []templates.TemplateDefinition
```
JacobsonMT: Might as well pre-size this.

JacobsonMT (approved):
Looks mostly good to me, left a couple comments and suggestions. They are all non-blocking except possibly the one about `result` and `a` sharing a backing array.

Will approve as-is though as the fixes are all straightforward.

definition/merge.go
yuri-tceretian: I think you mean `a`, right? If `b` has duplicates the resulting slice would contain deduped ones.

definition/merge_test.go
yuri-tceretian: we do not drop but rename it.

definition/merge.go
JacobsonMT: I meant `b` but I misunderstood the code, so the problem isn't as bad as I thought. It won't drop if there are multiple duplicates it will append numbers as needed.

Now the only problem is the `renamed` map won't be enough to identify which among the multiple duplicates were renamed. Not sure how necessary fixing this will be though, unless we're expecting duplicate named templates it might be easier just to validate their uniqueness.

definition/merge_test.go
JacobsonMT: You're right I misunderstood the code, it won't drop if there are multiple duplicates it will append numbers as needed.

Answered in more detail in the other comment, but basically consider what should be in the `renamed` map for this case:

```go
left: []PostableApiTemplate{{Name: "test", Kind: "grafana"}},
right: []PostableApiTemplate{{Name: "test", Kind: "grafana"}, {Name: "test", Kind: "grafana"}},
```

or even in the existing case. Which template was renamed, the first or the second?

definition/merge.go
yuri-tceretian: I decided to remove merge because it's general implementation does not make sense at this point when we have just two maps (grafana template and mimir ones) that guarantee uniqueness of the name. It will be much simpler to just convert maps to api models and concatenate slices.

PR #355, opened 2025-07-10
Enable Oauth2 for most http-based notifiers

JacobsonMT:
Followup to #338

This PR contains two commits:
- [Move http_config to a common area](https://github.com/grafana/alerting/commit/912bf07dabbdff31265cbc7dfa81509d12d174c0): This moves `http_config` into a more central area so it can be easily extended to other notifiers and their tests. `http_config` is now un-marshalled separate from the notifier-specific configuration.
- [Enable Oauth2 for most http-based notifiers](https://github.com/grafana/alerting/commit/c099b53a2748b669160b34aa08ed50894a47d168): This does the actual enabling on the common HTTP config options to most other notifiers. A much simpler PR because of the above prep.

Testing of the common HTTP config options (currently just oauth2) is fairly comprehensive and centralized in [notify/receivers_test.go](https://github.com/grafana/alerting/pull/355/files#diff-df280908f9627770e3f2f93e17b7e31aaedefbba9a166b03e02ccd35314a049eR275) instead of spread out among each notifier's package tests. The tests use mock servers to ensure the token is actually being correctly retrieved from the oauth2 server and the `Authorization` header set on outgoing http calls.

Notifiers not yet supported and why:
- Email: Choice of implementation isn't clear.
- Slack: Needs to use standard http client and handle Bot Token required fields if oauth2 is configured without a token.
- SNS: Needs to integrate with standard http client.
- MQTT: Implementation non-trivial.
- Alertmanager: Needs to use standard http client.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,8 @@
 module github.com/grafana/alerting
 
-go 1.23.0
+go 1.24.0
 
-toolchain go1.23.7
+toolchain go1.24.4
 
 require (
 	github.com/Masterminds/sprig/v3 v3.2.1
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,8 @@
 module github.com/grafana/alerting
 
-go 1.24.0
+go 1.23.0
 
-toolchain go1.24.4
+toolchain go1.23.7
 
 require (
 	github.com/Masterminds/sprig/v3 v3.2.1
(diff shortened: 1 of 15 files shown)


yuri-tceretian (approved):
LGTM
I am not a fan of the making the model generic for all integrations as there might be situations when particular implementation would require different settings and maintaining aggregated model would make it a nightmare. However, I understand the reason why you do it now, and I think it makes sense.

yuri-tceretian (approved):

PR #356, opened 2025-07-11
Bump go and golangci-lint version

JacobsonMT:
Bumping to go 1.24 breaks golangci-lint. So, the v1 .golangci.yml is migrated to v2 at a similar version to Grafana and various innocuous differences are fixed in code to make it pass.


yuri-tceretian (approved):
LGTM

yuri-tceretian (approved):

PR #363, opened 2025-08-01
Move notification historian to `grafana/alerting`

vstpme:
This PR moves multiple packages from `grafana/grafana` to `grafana/alerting` so they can be used by both Grafana and Mimir Alertmanager:
- `client` (HTTP client with metrics and traced, used by `lokiclient`)
- `lokiclient` (Loki client used by both state history and notification history)
- `notificationhistorian` (The actual implementation of notification history). Also addressed @yuri-tceretian's [comments](https://github.com/grafana/grafana/pull/107644#discussion_r2214298723).

Related PRs:
- https://github.com/grafana/grafana/pull/109078
- https://github.com/grafana/mimir/pull/12431


lokiclient/logproto/LICENSE_APACHE2
fayzal-g, line 1: Can we move import `github.com/grafana/loki` directly rather than copy and pasting the files from there?

notify/nfstatus/integration.go
```diff
@@ -112,6 +120,27 @@ func (n *statusCaptureNotifier) Notify(ctx context.Context, alerts ...*types.Ale
...
+		return
+	}
+	groupLabels, ok := notify.GroupLabels(ctx)
+	if !ok {
+		return
+	}
+	pipelineTime, ok := notify.Now(ctx)
+	if !ok {
+		return
+	}
```
fayzal-g: Would be good to have logs here in case of error as we don't return anything.

fayzal-g (commented):
Couple small comments

lokiclient/logproto/LICENSE_APACHE2
vstpme: good call! done in https://github.com/grafana/alerting/pull/363/commits/4f8a1edf8f5f5cd07b0184103b604a0e8a000cd9

notify/nfstatus/integration.go
vstpme: done in https://github.com/grafana/alerting/pull/363/commits/bf72d2bdfd7e6a4ec8a2219e90681bc9430093e0

client/client.go
```diff
@@ -0,0 +1,125 @@
+package client
```
yuri-tceretian: This is too generic, especially in the context of the module and being the top level package. This needs a better naming. 
The package seems to implement two wrappers that let us instrument HTTP requests with metrics and traces. 
I think it could go to `http/instrument`

lokiclient/testing.go
```diff
@@ -0,0 +1,45 @@
+package lokiclient
+
+import (
+	"bytes"
+	"io"
+	"net/http"
+)
+
+type FakeRequester struct {
```
yuri-tceretian: Considering my proposal for moving client to `http/instrument`, I think these fakes could go into  `http/instrument/instrumenttest` This will make the fake reusable in other places that happen to import instrumented clients

client/client.go
```diff
@@ -0,0 +1,125 @@
...
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/propagation"
+	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// Requester executes an HTTP request.
+type Requester interface {
+	Do(req *http.Request) (*http.Response, error)
+}
```
yuri-tceretian: why do we need this interface that just copies http.Client? Perhaps let's use the standard one?

lokiclient/client.go
```diff
@@ -0,0 +1,292 @@
...
+	Eq Operator = "="
+	// Not Equal operator (!=)
+	Neq Operator = "!="
+	// Equal operator supporting RegEx (=~)
+	EqRegEx Operator = "=~"
+	// Not Equal operator supporting RegEx (!~)
+	NeqRegEx Operator = "!~"
+)
+
+func NewLokiClient(cfg LokiConfig, req client.Requester, bytesWritten prometheus.Counter, writeDuration *instrument.HistogramCollector, logger log.Logger, tracer trace.Tracer, spanName string) *HTTPLokiClient {
```
yuri-tceretian: I think we need to get rid of this client and vendor https://github.com/grafana/loki/blob/main/pkg/logcli/client/client.go instead.

notify/nfstatus/integration.go
```diff
@@ -5,13 +5,25 @@ import (
...
+	Record(
+		ctx context.Context,
+		alerts []*types.Alert,
+		retry bool,
+		notificationErr error,
+		duration time.Duration,
+		receiverName string,
+		groupLabels model.LabelSet,
+		pipelineTime time.Time,
+		now time.Time,
```
yuri-tceretian: Let's move it to a struct. I bet in the future we will try to stick more info to it.

notify/nfstatus/integration.go
```diff
@@ -112,6 +123,30 @@ func (n *statusCaptureNotifier) Notify(ctx context.Context, alerts ...*types.Ale
...
+	if !ok {
+		level.Warn(n.logger).Log("msg", "failed to get group labels from context, skipping notification history write")
+		return
+	}
+	pipelineTime, ok := notify.Now(ctx)
+	if !ok {
+		level.Warn(n.logger).Log("msg", "failed to get pipeline time from context, skipping notification history write")
+		return
+	}
+	now := time.Now()
```
yuri-tceretian: I am not sure what's the reason for this time here. It does not reflect the time of the event precisely because it's called async and after who knows how many attempts. It looks like an implementation detail of loki client that requires the timestamp of the event.

notificationhistorian/historian.go
```diff
@@ -0,0 +1,222 @@
...
+	alerts []*types.Alert,
+	retry bool,
+	notificationErr error,
+	duration time.Duration,
+	receiverName string,
+	groupLabels prometheusModel.LabelSet,
+	pipelineTime time.Time,
+	now time.Time,
+) {
+	streams, err := h.prepareStreams(alerts, retry, notificationErr, duration, receiverName, groupLabels, pipelineTime, now)
```
yuri-tceretian: I would set now here and do not pass from the outside

notify/factory.go
```diff
@@ -250,7 +250,7 @@ func BuildPrometheusReceiverIntegrations(
 			if wrapper != nil {
 				n = wrapper(name, n)
 			}
-			integrations = append(integrations, nfstatus.NewIntegration(n, rs, name, i, nc.Name, notificationHistorian))
+			integrations = append(integrations, nfstatus.NewIntegration(n, rs, name, i, nc.Name, notificationHistorian, logger))
```
yuri-tceretian: let's use the logger from the line 245, at least it will give us better correlation in logs

go.mod
```diff
@@ -10,21 +10,27 @@ require (
...
-	github.com/davecgh/go-spew v1.1.1
+	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
 	github.com/go-kit/log v0.2.1
 	github.com/go-openapi/strfmt v0.22.0
-	github.com/google/go-cmp v0.6.0
+	github.com/gogo/protobuf v1.3.2
+	github.com/golang/snappy v0.0.4
+	github.com/google/go-cmp v0.7.0
+	github.com/grafana/dskit v0.0.0-20250723143816-ff33c5829b96
+	github.com/grafana/loki v1.6.1
```
yuri-tceretian: This does not seem to be required. Let's replace to module github.com/grafana/loki/pkg/push

lokiclient/encode.go
```diff
@@ -0,0 +1,104 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"slices"
+	"strconv"
+	"strings"
+
+	"github.com/gogo/protobuf/proto"
+	"github.com/golang/snappy"
+	"github.com/grafana/loki/pkg/logproto"
```
yuri-tceretian: let's replace to module `github.com/grafana/loki/pkg/push` logproto just aliases types from that module. This will let us avoid importing whole loki module into the project

yuri-tceretian (commented):
Overall, LGTM. 
However, I would recommend changing the structure of the packages introduced to the module.
- move `client` to `http/instrument`
- move notificationhistorian to `notify/historian`
- move lokiclient to `notify/historian/loki/client`

I do not really like the idea of bringing so many tracing dependencies here. If it wasn't for reusing it in Mimir alertmanager

client/client.go
vstpme: good point! moved to `http/instrument` in https://github.com/grafana/alerting/pull/363/commits/174784ef97fd6e3b92a9f68756e3e80ec138faae

lokiclient/testing.go
vstpme: done in https://github.com/grafana/alerting/pull/363/commits/4d0d5e93d975e3254b1ec72edd60eae4f4996873

notify/nfstatus/integration.go
vstpme: make sense! done in https://github.com/grafana/alerting/pull/363/commits/b6e84722c45a2cdbc9ffbad06533e9809dfc4310

notify/nfstatus/integration.go
vstpme: done in https://github.com/grafana/alerting/pull/363/commits/b6e84722c45a2cdbc9ffbad06533e9809dfc4310

notificationhistorian/historian.go
vstpme: done in https://github.com/grafana/alerting/pull/363/commits/b6e84722c45a2cdbc9ffbad06533e9809dfc4310

notify/factory.go
vstpme: done in https://github.com/grafana/alerting/pull/363/commits/03412f26a6524f521a313a8c9708ed97bfaee9d8

go.mod
vstpme: done in https://github.com/grafana/alerting/pull/363/commits/42f16d74b4ab494e57852595193877b8b1c17984

lokiclient/encode.go
vstpme: done in https://github.com/grafana/alerting/pull/363/commits/42f16d74b4ab494e57852595193877b8b1c17984

lokiclient/client.go
vstpme: Unfortunately that client only implements querying. Pushing is not implemented there.

client/client.go
vstpme: AFAICT it's mostly useful for stuff like this: https://github.com/grafana/alerting/blob/174784ef97fd6e3b92a9f68756e3e80ec138faae/lokiclient/client.go#L73-L74

Having `Requester` as a separate interface makes it possible to use `TimedClient` there which is not an `http.Client` but behaves like one.

vstpme:
@yuri-tceretian thank you for the review! In addition to my comments above, I also moved the packages around as you suggested:
* `notificationhistorian` -> `notify/historian`
* `lokiclient` -> `notify/historian/lokiclient`

This was done in https://github.com/grafana/alerting/pull/363/commits/6e3c2d708b8d461019de5e100a908c68f0887992

client/client.go
yuri-tceretian: ah, you're right! Due to some reason I thought http.Client is an interface, but it's a struct

yuri-tceretian (approved):
LGTM! 👍🏻

fayzal-g (approved):

PR #364, opened 2025-08-08
Add ability to omit some templates from builtin definition

yuri-tceretian:
The built-in template definition contains email templates that in fact are not used in Grafana. Instead, they confuse users by showing up in the drop down of field templates, and when they try to use it, the resulting notification arrives with broken formatting, which makes users think this is a bug.

I chose to make it as parameter, so it will be clearer on the Grafana side that the resulting definition is not full.


rwwiv (approved):
Looks like there's minor a lint error in the test, but the code change LGTM

PR #369, opened 2025-08-21
Move Jira search to a new endpoint

yuri-tceretian:
This PR changes the JIRA integration to use a different search endpoint. 
The current `/search` endpoint got deprecated (see https://developer.atlassian.com/changelog/#CHANGE-2046 ). Instead, this PR proposes to use `/search/jql` (see https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-rest-api-3-search-jql-post), which mostly compatible with the old one except a few fields.


JacobsonMT (approved):
LGTM

PR #379, opened 2025-09-10
Move notification history to grafana/alerting

vstpme:
Reopening https://github.com/grafana/alerting/pull/363 after reverting it, but with the following changes to `go.mod`:
- `github.com/grafana/loki/pkg/push`: use `v0.0.0-20250823105456-332df2b20000` instead of latest
- `github.com/grafana/dskit`: use `v0.0.0-20250818234656-8ff9c6532e85` instead of latest

These changes are made to avoid bumping `github.com/stretchr/testify` from `v1.10.0` to `v1.11.1`, which breaks some random tests in `grafana/grafana`: https://github.com/grafana/grafana/pull/109078


notify/historian/historian.go
```diff
@@ -0,0 +1,204 @@
...
+		Values: []lokiclient.Sample{
+			{
+				T: now,
+				V: string(entryJSON),
+			}},
+	}, nil
+}
+
+func prepareLabels(labels prometheusModel.LabelSet) map[string]string {
+	result := make(map[string]string)
```
yuri-tceretian: I think we can specify capacity here

notify/historian/historian.go
```diff
@@ -0,0 +1,204 @@
...
+		level.Error(h.logger).Log("msg", "Failed to convert notification history to streams", "error", err)
+		return
+	}
+
+	// This is a new background job, so let's create a new context for it.
+	// We want it to be isolated, i.e. we don't want grafana shutdowns to interrupt this work
+	// immediately but rather try to flush writes.
+	// This also prevents timeouts or other lingering objects (like transactions) from being
+	// incorrectly propagated here from other areas.
+	writeCtx, cancel := context.WithTimeout(context.Background(), NotificationHistoryWriteTimeout)
```
yuri-tceretian, line 98: I am not sure whether this is perfect idea, to be honest. This approach does not guarantee the request is sent when the process is getting stopped. To do it properly, we need to register historian as a background service (in Grafana), with Start\Stop semantic and support for graceful shutdown.

notify/historian/historian.go
```diff
@@ -0,0 +1,204 @@
...
+	writeCtx = trace.ContextWithSpan(writeCtx, trace.SpanFromContext(ctx))
```
yuri-tceretian, line 99: this can cause weird situations when span gets closed and then we add more events to it.

yuri-tceretian (approved):
LGTM

notify/historian/historian.go
vstpme: done in https://github.com/grafana/alerting/pull/379/commits/e92c68a5e6f53055284d35e59676b51dad9d4929

notify/historian/historian.go
vstpme: yeah that makes sense.. I copied this from state history, but probably there are betters ways to do it.

this alternative could work (according to Claude Code), but I'm not 100% sure:
```golang
writeCtx := trace.ContextWithSpanContext(context.Background(), trace.SpanContextFromContext(ctx))
writeCtx, cancel := context.WithTimeout(writeCtx, NotificationHistoryWriteTimeout)
defer cancel()
```

I'll do more research on this and address it in a separate PR.

notify/historian/historian.go
vstpme: same as https://github.com/grafana/alerting/pull/379#discussion_r2344126326

PR #384, opened 2025-09-17
Integration schema

yuri-tceretian:
This PR introduces schema to receivers package. The schema provides a detailed description a integration configuration, available fields and how they are supposed to be presented in user interface.  The schema is a copy of [plugin.go](https://github.com/grafana/grafana/blob/c36b2ae191550a0110937f95dda6795a912487f6/pkg/services/ngalert/notifier/channels_config/plugin.go) from Grafana repository. 
Also, this PR refactors receivers package and moves all current integrations configurations into v1 sub-package. This will reflect the future structure of versioned integrations. 

### Why we need this?
Currently, the schema lives in Grafana and is served by its API. This creates a couple of difficulties:
- when a new integration is created or an existing one is modified, the change needs to be done in a multiple places, and sometimes contributors miss some. Bringing schema into the same place as the configuration definition makes simpler.
- in multiservice mode, when Alertmanager runs outside the Grafana service, there could be a discrepancy in integration version  supported by Alertmanager and Grafana. Moving schema into alerting module will let us make Alertmanager serve the schema and therefore eliminate this problem.

### Notes for reviewers

All integration schemas are copied from the [current](https://github.com/grafana/grafana/blob/c36b2ae191550a0110937f95dda6795a912487f6/pkg/services/ngalert/notifier/channels_config/available_channels.go) schema in Grafana


notify/factory.go
```diff
@@ -34,9 +34,9 @@ import (
 	"github.com/grafana/alerting/images"
 	"github.com/grafana/alerting/notify/nfstatus"
 	"github.com/grafana/alerting/receivers"
-	"github.com/grafana/alerting/receivers/alertmanager"
-	"github.com/grafana/alerting/receivers/dinding"
-	"github.com/grafana/alerting/receivers/discord"
+	alertmanager "github.com/grafana/alerting/receivers/alertmanager/v1"
+	dingding "github.com/grafana/alerting/receivers/dingding/v1"
+	discord "github.com/grafana/alerting/receivers/discord/v1"
```
titolins, line 39: An alternative would be to start tagging releases and using go module versioning so there's no need to rename imports/packages at this point. Renaming might be simpler though 👍

titolins (commented):
Nice! I've just scrolled by quickly for now. It looks good but I'll go over it again with more time by tomorrow 👍

notify/factory.go
yuri-tceretian: v1 is for Grafana integrations. `v0*` packages will be dedicated to mimir integraitons.

moustafab (approved):
This looks great and is a useful consolidation!

My only thoughts while reviewing this were that it feels like the `config.go` files almost get in the way of what seems like it could be expressed in a bit of an easier to read and understand way for a human in a different format. Perhaps some code generation would be nice here at some point to cut down on the boiler plate?

yuri-tceretian:
>My only thoughts while reviewing this were that it feels like the config.go files almost get in the way of what seems like it could be expressed in a bit of an easier to read and understand way for a human in a different format. Perhaps some code generation would be nice here at some point to cut down on the boiler plate?

Totally agree! I was thinking about adding tags to the config files and then generate the schema on fly.

PR #385, opened 2025-09-18
Disallow matchers and object matchers in root route

moustafab:
This is done in [upstream](https://github.com/prometheus/alertmanager/blob/17f42810382299639b34b56a01550243f2b2512a/config/config.go#L647), and we're doing the same for match and match regex, but we missed disallowing object matchers and matchers which replace them.
If we allow them very confusing behavior can happen, where notifications are dropped because the root route has matchers that don't match.

Note: This may cause issues where previously "valid" configurations are no longer valid. We will need to provide guidance on handling those cases, but basically in order to maintain the behavior the "correct" way of ignoring notifications is to use a null receiver.


yuri-tceretian (approved):
LGTM

PR #391, opened 2025-09-29
Move common structs from notify to models package

yuri-tceretian:
This PR is a minor refactoring:
- Moves `notify.GrafanaIntegrationConfig` to `models.IntegrationConfig`
- Moves `notify.GrafanaIntegrations` to `models.ReceiverConfig`
- Moves compat functions [from Grafana ](https://github.com/grafana/grafana/blob/5520f481538f5dd4660a05127fb7b25ba0d96c9b/pkg/services/ngalert/notifier/compat.go#L12-L43) to notify
- Renames models.Receiver and models.Integration to ReceiverStatus and IntegrationStatus to better reflect thier purpose.

This will make it reusable and avoid some circular dependencies

### DIFF
--- a/models/integration.go
+++ b/models/integration.go
@@ -10,3 +10,7 @@ type IntegrationConfig struct {
 	Settings              json.RawMessage   `json:"settings" yaml:"settings"`
 	SecureSettings        map[string]string `json:"secureSettings" yaml:"secureSettings"`
 }
+
+type ReceiverConfig struct {
+	Integrations []*IntegrationConfig `yaml:"grafana_managed_receiver_configs,omitempty" json:"grafana_managed_receiver_configs,omitempty"`
+}
--- a/models/integration.go
+++ b/models/integration.go
@@ -1,0 +1,12 @@
+package models
+
+import "encoding/json"
+
+type IntegrationConfig struct {
+	UID                   string            `json:"uid" yaml:"uid"`
+	Name                  string            `json:"name" yaml:"name"`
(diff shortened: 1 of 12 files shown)


models/integration.go
```diff
@@ -0,0 +1,16 @@
...
+type IntegrationConfig struct {
+	UID                   string            `json:"uid" yaml:"uid"`
+	Name                  string            `json:"name" yaml:"name"`
+	Type                  string            `json:"type" yaml:"type"`
+	DisableResolveMessage bool              `json:"disableResolveMessage" yaml:"disableResolveMessage"`
+	Settings              json.RawMessage   `json:"settings" yaml:"settings"`
+	SecureSettings        map[string]string `json:"secureSettings" yaml:"secureSettings"`
+}
+
+type ReceiverConfig struct {
```
yuri-tceretian, line 14: No name here because we embed it into APIReceiver where it would collide with field Name from the upstream. Later we can wrap into Receiver struct

notify/testing.go
```diff
@@ -118,37 +115,3 @@ func MergeSettings(a []byte, b []byte) ([]byte, error) {
 
 	return json.Marshal(origSettings)
 }
-
-func GetRawNotifierConfig(n notifytest.NotifierConfigTest, name string) *GrafanaIntegrationConfig {
```
yuri-tceretian, line 122: This reverts the trade-off I had to make in https://github.com/grafana/alerting/pull/389.

JacobsonMT (approved):
LGTM

PR #392, opened 2025-10-01
Improve schemas

yuri-tceretian:
This PR makes the schema design more idiomatic. 
- Make all schema package variables
- Initialize all schemas once to a map for faster access. (schema is read pretty often in receivers API)
- Adds reference to IntegrationTypeSchema to IntegrationSchemaVersion. This will let us pass only the latter throughout the code and always have information of the type.
- Introduces schema validation and test to ensure all known schema pass them
- Copies methods to work with secure fields from Grafana's models.IntegrationConfig including type IntegrationFieldPath
- Adds more functions to access schema


notify/schema.go
```diff
@@ -30,52 +33,184 @@ import (
...
+		wechat.Schema,
+		wecom.Schema,
+	}
+
+	allSch := make(map[schema.IntegrationType]schema.IntegrationTypeSchema, len(all))
+	aliases := make(map[schema.IntegrationType]schema.IntegrationType)
+	for _, sch := range all {
+		if _, ok := allSch[sch.Type]; ok {
+			panic(fmt.Sprintf("duplicate schema type %s", sch.Type))
+		}
```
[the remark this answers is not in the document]
yuri-tceretian: It should be ok because it can only happen only when someone adds a new type or alias. This panic is only possible in CI, and it will be clear where it comes from

receivers/schema/schema.go
JacobsonMT, line 1: There seems to be some modifications to the code copied over from `grafana/grafana`, some of which I believe contains regressions. For example:

- `traverseFields` appends to the `parentPath` directly again, this was previously fixed here: https://github.com/grafana/grafana/pull/106302/files#diff-fcb376db2dd25afc74032d23bd52315d5c65fd9b3c16bae53e1a309cb51ef0cb.

Also, the test that covers this case appears to have been lose during this migration.

yuri-tceretian: Do you mean this https://github.com/grafana/alerting/pull/392/files#diff-de8d1bde76c3bdfaa261a5a220ca537ee88f33a7958cb8656ed64fd44198f068R161-R171
Yea, I modified it to not copy the slice in all the cases while it traverses non-secure fields. It will create a copy via "With" method only if it finds secure field.
I covered it with some tests [here](https://github.com/grafana/alerting/pull/392/files) + [one](https://github.com/grafana/alerting/blob/38e3d5cc6709fa63e605643e2defd11953053f71/notify/schema_test.go#L45) that asserts all known secrets

JacobsonMT (approved):

PR #401, opened 2025-10-06
Convert config.Receiver to a slice of integration configs

yuri-tceretian:
1. Adds a functions to resolve v0 integration types from struct field types
`IntegrationTypeFromMimirType` and `IntegrationTypeFromMimirTypeReflect`. They help convert a field into `schema.IntegrationType`.

2. Adds a adapter function to convert Alertmanager's Reciever into a generic struct `MimirIntegrationConfig` that holds information about type and version along with configuration. 

This will help us convert v0 configurations to models we can work in a generic manner

### DIFF
--- a/notify/compat.go
+++ b/notify/compat.go
@@ -109,7 +109,13 @@ func ConfigReceiverToIntegrations(receiver ConfigReceiver) ([]MimirIntegrationCo
 		}
 		result = slices.Grow(result, len(result)+sliceVal.Len())
 		for j := 0; j < sliceVal.Len(); j++ {
-			elem := sliceVal.Index(j).Elem().Interface()
+			var elem any
+			item := sliceVal.Index(j)
+			if elemType.Kind() == reflect.Ptr {
+				elem = item.Elem().Interface()
+			} else {
+				elem = item.Interface()
+			}
 			result = append(result, MimirIntegrationConfig{
 				Schema: version,
 				Config: elem,
--- a/notify/compat.go
+++ b/notify/compat.go
@@ -87,6 +87,10 @@ func ConfigReceiverToIntegrations(receiver ConfigReceiver) ([]MimirIntegrationCo
 			continue
 		}
 		iType, err := IntegrationTypeFromMimirTypeReflect(elemType)
+		if err != nil {
+			return nil, err
+		}
+
 		sch, ok := GetSchemaForIntegration(iType)
 		if !ok {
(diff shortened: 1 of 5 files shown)


notify/compat.go
```diff
@@ -65,3 +67,35 @@ func PostableAPITemplatesToTemplateDefinitions(ts []definition.PostableApiTempla
...
+		if sliceVal.Len() == 0 {
+			continue
+		}
+		iType, err := IntegrationTypeFromMimirTypeReflect(elemType)
+		if err != nil {
+			return nil, err
+		}
+		slices.Grow(result, sliceVal.Len())
+		for j := 0; j < sliceVal.Len(); j++ {
+			elem := sliceVal.Index(j).Elem().Interface()
```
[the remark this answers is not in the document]
yuri-tceretian: All slices of the type ConfigReceiver contain pointers. So this should be a pretty safe assumption but I think it makes sense to do it to not panic in case of it changes in the future

notify/compat.go
```diff
@@ -65,3 +69,58 @@ func PostableAPITemplatesToTemplateDefinitions(ts []definition.PostableApiTempla
...
+			if !ok {
+				return nil, fmt.Errorf(" integration type %s does not have version %s", iType, schema.V0mimir1)
+			}
+		} else {
+			version, ok = sch.GetVersionByTypeAlias(iType)
+			if !ok {
+				return nil, fmt.Errorf("cannot find schema version by integration type alias %s", iType)
+			}
+		}
+		result = slices.Grow(result, len(result)+sliceVal.Len())
```
JacobsonMT: This should be the following, no?
```suggestion
		result = slices.Grow(result, sliceVal.Len())
```

notify/compat.go
```diff
@@ -65,3 +69,58 @@ func PostableAPITemplatesToTemplateDefinitions(ts []definition.PostableApiTempla
 	}
 	return defs
 }
+
+// ConfigReceiverToIntegrations converts a ConfigReceiver to a list of MimirIntegrationConfig
+func ConfigReceiverToIntegrations(receiver ConfigReceiver) ([]MimirIntegrationConfig, error) {
```
JacobsonMT: WDYT: `ConfigReceiverToMimirIntegrations`

JacobsonMT (approved):
LGTM with one non-blocking issue

notify/compat.go
yuri-tceretian: ah, you're right.

PR #402, opened 2025-10-08
Make MergeReceivers and MergeTimeIntervals public

yuri-tceretian:
- Make MergeReceivers and MergeTimeIntervals public functions
- Fixes MergeReceivers to not modify incoming lists of receivers
- Adds tests for both functions 
- Updates MergeTimeIntervals to convert mute time intervals into time intervals preserving order of precedence that is used during the configuration of the Alertmanager

### DIFF
--- a/definition/merge.go
+++ b/definition/merge.go
@@ -190,43 +190,48 @@ func MergeTimeIntervals(
 	incomingTimeIntervals []config.TimeInterval,
 	suffix string,
 ) ([]config.TimeInterval, map[string]string) {
-	usedNames := make(map[string]int, len(existingMuteIntervals)+len(incomingTimeIntervals)+len(incomingMuteIntervals)+len(existingTimeIntervals))
+	// combine all incoming intervals into a single list
+	incomingAll := make([]config.TimeInterval, 0, len(incomingTimeIntervals)+len(incomingMuteIntervals))
+	incomingAll = append(incomingAll, incomingTimeIntervals...)
+	for _, interval := range incomingMuteIntervals {
+		incomingAll = append(incomingAll, config.TimeInterval(interval))
+	}
+	usedNames := createIndexTimeIntervals(existingMuteIntervals, existingTimeIntervals, incomingAll)
+	result := make([]config.TimeInterval, 0, len(existingTimeIntervals)+len(incomingMuteIntervals)+len(incomingTimeIntervals))
+	result = append(result, existingTimeIntervals...)
+	renames := make(map[string]string)
+	for idx, interval := range incomingAll {
+		curName := interval.Name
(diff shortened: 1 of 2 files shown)


definition/merge.go
```diff
@@ -181,32 +181,52 @@ func Merge(a, b PostableApiAlertingConfig, opts MergeOpts) (MergeResult, error)
...
+// It returns a merged list of time intervals and a map of renamed interval names for tracking adjustments made. Mute time intervals are converted to time intervals.
+func MergeTimeIntervals(
+	existingMuteIntervals []config.MuteTimeInterval,
+	existingTimeIntervals []config.TimeInterval,
+	incomingMuteIntervals []config.MuteTimeInterval,
+	incomingTimeIntervals []config.TimeInterval,
+	suffix string,
+) ([]config.TimeInterval, map[string]string) {
+	usedNames := make(map[string]int, len(existingMuteIntervals)+len(incomingTimeIntervals)+len(incomingMuteIntervals)+len(existingTimeIntervals))
+	for _, r := range existingMuteIntervals {
```
JacobsonMT: I find it a bit hard to follow the key aspects of this function the way it's laid out. For example, the order preservation is peppered throughout the function and renaming relies on a stored map index which I'm not sure is needed.

WDYT about using a helper function like:

```go
	used := make(map[string]struct{}, len(existingMuteIntervals)+len(existingTimeIntervals)+len(incomingMuteIntervals)+len(incomingTimeIntervals))
	renames := make(map[string]string)
	result := make([]config.TimeInterval, 0, len(existingTimeIntervals)+len(incomingTimeIntervals)+len(incomingMuteIntervals))

	appendInterval := func(ti config.TimeInterval) {
		originalName := ti.Name
		if _, exists := used[ti.Name]; exists {
			ti.Name = getUniqueName(originalName, suffix, used)
			renames[originalName] = ti.Name
		}

		result = append(result, ti)
		used[ti.Name] = struct{}{}
	}
```

This should let all the business logic to be short and sweet at the end:

```go
	for _, mi := range existingMuteIntervals {
		used[mi.Name] = struct{}{}
	}
	for _, ti := range existingTimeIntervals {
		appendInterval(ti)
	}
	for _, ti := range incomingTimeIntervals {
		appendInterval(ti)
	}
	for _, mi := range incomingMuteIntervals {
		appendInterval(config.TimeInterval(mi))
	}
```

yuri-tceretian: Yea, I agree the readability could be improved. I will see what I can do. 

In this case prepping the map is necessary to avoid extra renames, e.g. case when 
existing: 
- ti1

incoming:
- ti1
- ti1-dedup

There is a test that covers this aspect.

JacobsonMT (approved):
LGTM

PR #406, opened 2025-10-22
Integration testing

yuri-tceretian:
This PR introduces a method to test only one integration. In contrast to the existing method that offers to test multiple receivers, which in fact is not used in UI, the new one implements a very simple interface. 

Apart from the new method, there is some minor refactoring to reuse existing components

related to https://github.com/grafana/grafana/issues/112823


alexander-akhmetov (approved):

notify/test_receivers.go
```diff
@@ -0,0 +1,71 @@
+package notify
```
titolins, line 1: just a nit, but I always find it a bit confusing when we have `..._testing.go` or somewhat similar pattern to what go expects as a test file.. maybe `test_(integration|receiver).go` would be more better; wdyt?

titolins (approved):

notify/test_receivers.go
yuri-tceretian: that makes sense!

PR #416, opened 2025-11-03
feat(receivers/jira): adding templating to custom fields

rcmi26133:
Currently, templatization only works on select jira fields.  This PR adds templating to custom fields.


yuri-tceretian (approved):
LGTM! Thanks for the contribution.

PR #419, opened 2025-11-19
alertmanager: add dispatch timer opts

titolins:
## Context
- Adding support for sync timer so we can start rolling it out
- Sync timer synchronizes the flush times across HA nodes to decrease the chance of duplicate notifications
- If opt timer is set to sync timer, we initialize the flush log and set the sync timer factory when initializing the dispatcher

Related to https://github.com/grafana/grafana/pull/114602


yuri-tceretian (approved):
LGTM

PR #427, opened 2025-11-28
Alerting: Migrate Slack to use Block Kit

TimShilov:
> [!NOTE]
> This is a rough draft, just to start the conversation and get some guidance. It's not polished and not well tested yet.
> I'd like to know first, **is it something worth finalizing and would maintainers consider merging it if I'll do so?**

In this PR I attempted to migrate existing Slack receiver to use [Block Kit](https://docs.slack.dev/block-kit/) instead of the [legacy attachments](https://docs.slack.dev/legacy/legacy-messaging/legacy-secondary-message-attachments/) (which are [discouraged from using by the Slack documentation](https://docs.slack.dev/messaging/formatting-message-text#when-to-use-attachments)).
I have migrated it according to [the official migration guide](https://docs.slack.dev/messaging/migrating-outmoded-message-compositions-to-blocks).

In this initial migration PR I've tried not to change the way notifications actually look at the moment.
<details><summary><strong>Here's the example of how test alert looked like BEFORE and AFTER the change:</strong></summary>
<p>BEFORE:</p>
Before

<p>AFTER:</p>
After
</details> 

The only visual differences between BEFORE and AFTER are:
- Spacing between title and message is a bit different (arguably, it is better than before)
- The icon in the footer looks a bit bigger and is better aligned
- The timestamp that said "Today at 16:31" is missing in the footer. It is possible to replicate using [date formatting](https://docs.slack.dev/messaging/formatting-message-text#date-formatting), though I couldn't find the exact format AND I really don't see the point in having this timestamp to begin with given that the message itself has the timestamp already. 🤷‍♂️ 

---
Hopefully, if this will be merged, then it will be possible to actually improve the layout of the Slack notifications to make them look better, and maybe even try to address https://github.com/grafana/grafana/issues/82843 and https://github.com/grafana/grafana/issues/80374 somehow.

Questions/problems to address:
1. There's one problem I haven't found the solution to - the `fallback` field. The documentation suggests to use top-level `text` instead and it works just fine but ONLY if there's anything in top-level `blocks` as well. I've put mentions in that, but if Contact Point isn't configured to mention anyone - then the `blocks` is empty and `text` is actually displayed. So the only way to workaround this is to always have something inside top-level `blocks`. It might be a good idea to put the `title` outside of `attachments` but that will look very different from the current notifications.

<details><summary>Here's what that could look like</summary>
<p>Using markdown</p>
Screenshot 2025-11-28 at 17 04 23

<p>Using header</p>
Screenshot 2025-11-28 at 17 06 44

</details> 

2. Apparently, Slack doesn't recommend using `attachments` field at all, but it seems to be the only way to create the colored border. I wonder if we should consider getting rid of the colored border? That not only would allow us to stop using deprecated field, but also would solve the above-mentioned issue with `fallback`... but the color looks nice. 😬


YashIngole:
Any update on this PR?

PR #436, opened 2025-12-08
Improve alerting-gen

santihernandezc:
This PR introduces some improvements to the `alerting-gen` module:

1. Concurrency for Performance
    - Added `-c` flag (default 10) to control concurrent requests to the provisioning API
    - Significantly faster rule provisioning for large-scale scenarios
2. Dynamic Folder Management
    - Added `--num-folders` flag to create folders dynamically via API
    - Added `--nuke` flag to delete all alerting-gen created folders (useful for testing/cleanup)
    - Folders are identified by _"Alerts Folder"_ prefix for deletion
3. Configuration Defaults & Options
    - Changed `--alerts` default from 10 to 0 (opt-in behavior)
    - Changed `--query-ds` default from `__expr__` to `grafanacloud-prom`
    - Made `--write-ds` default to `--query-ds` value (same data source is used for reading and writing metrics in recording rules)
    - Added `--eval-interval` flag to explicitly set evaluation intervals (otherwise random 1-20m)
4. Authentication
    - Removed custom transport in favor of Grafana OpenAPI client's built-in auth mechanisms
5. Default Prometheus Queries
    - Changed the default query from `rate(http_requests_total[5m])` (no data) to `vector(1)`
6. Reliability Improvements
    - Increased HTTP client retries to 100 with 1-minute timeout
    - Better error messages throughout
    - Only marshals JSON when output file is specified
  7. Misc
     - Used min() built-in function, shorter `range` syntax, fixed comment punctuation, etc

_Note: We would benefit from validating the config, but I'd prefer to do that in a separate PR._


testing/alerting-gen/pkg/auth/transport.go
```diff
@@ -1,49 +0,0 @@
...
-		base = http.DefaultTransport
-	}
-	return &BasicAuthRoundTripper{base: base, username: username, password: password}
-}
-
-func (b *BasicAuthRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
-	creds := b.username + ":" + b.password
-	req.Header.Set("Authorization", "Basic "+base64.StdEncoding.EncodeToString([]byte(creds)))
-	return b.base.RoundTrip(req)
-}
```
santihernandezc, line 49: There's an easier way to add these headers to the client

alexander-akhmetov (approved):

PR #437, opened 2025-12-10
Add limit to template expansion

yuri-tceretian:
This PR 
- Intoduces an implementation of `io.Writer` that allows to limit the amount of data written.
- Limits the output of the template expansion in template testing API to 1MB per expansion.
- Limits the output of the template expansion in notification workflow to 10Mb per expansion.

Related to https://github.com/grafana/alerting-squad/issues/1327

PR #438, opened 2025-12-11
Configurable template output limits

yuri-tceretian:
This PR builds on the feature https://github.com/grafana/alerting/pull/437 and make the template output limits be propagated from the configuration. 
Key changes:
- `templates.Template` is not an alias to upstream's Template but an extension of it. 
- A few new structs are introduced in templates package: Config and Limits 
- DispatchLimit field in NotificationConfiguration is replaced with DynamicLimits that include dispatch and template limits

NOTE: the hardcoded limit for template testing is not changed and is still 1Mb

### DIFF
--- a/templates/factory_test.go
+++ b/templates/factory_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"maps"
+	"net/url"
 	"testing"
 	"time"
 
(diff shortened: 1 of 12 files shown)


templates/factory_test.go
```diff
@@ -53,24 +54,20 @@ func TestNewFactory(t *testing.T) {
...
 			for kind, expectedTemplates := range tc.expected {
 				require.ElementsMatch(t, expectedTemplates, factory.templates[kind])
 			}
 		})
 	}
-
-	t.Run("error if external URL is invalid", func(t *testing.T) {
-		_, err := NewFactory([]TemplateDefinition{{Name: "t1", Kind: GrafanaKind}}, logger, ":::", "grafana")
-		require.Error(t, err)
-	})
```
alexander-akhmetov, line 73: Should it be replaced with a test for NewConfig that verifies it returns an error for invalid URLs?

alexander-akhmetov (approved):

templates/factory_test.go
yuri-tceretian: yes, there are bunch of tests

PR #448, opened 2025-12-29
grafana/grafana#89393 telegram 2-step send to fix formatting errors

### LINKED ISSUE #89393: Truncating Telegram HTML/Markdown alerts returns 400
Up to recently I thought that Grafana did not truncate messages to 4096 and that was the reason I kept getting errors 400 from Telegram API (in the logs).

When considering doing a PR I found https://github.com/grafana/alerting/blob/main/receivers/telegram/telegram.go#L120 and tested simpler alerts and indeed it was truncated and alert sent successfully.

Then I found somewhere that Telegram requires messages with Markdown/HTML parse_mode to be *valid* and I finally managed to find my issue: truncating often removes a terminating tag (as I use HTML) leaving the opening tag, resulting in invalid HTML.

I've reproduced this easily using `Contact points` test message.
I just filled a message `<b>a</b>` preceeded by 4089 `A` and clicking `Test` results in error 400.

Error 400s are easily confirmed by just using `<b>a` as message or `_a` with Markdown parse mode.

I usually like to provide possible solutions yet I don't have an easy one in this case 😞 

I believe the proper way would be to, after truncating, to loosely parse the message (in the proper parse_mode) and close any blocks that are missing, yet "sounds heavy"...

A simpler workaround could be to a new feature which is to use allow a `Contact Point` to have "fallback" integrations.
Currently one with multiple integrations will always notify all integrations. If there was a "fallback" integration, it would mean it's only notified if a default one fails.
This would allow to set up a contact point with 2 Telegram integrations: one with pretty markdown/html template and a second with a simpler template and parse_mode `None`.

danilgotvyansky:
## Description:

Here is how I tried to fix the issue with Grafana and Telegram communication being corrupted by Telegram failing to validate parsing syntaxes if the message is truncated

I am not a Go developer, but I've tried to show my idea:
- Format message **after** the truncation to get guaranteed delivery with `none` parse mode
- To achieve this, we send a message with `none` parse mode always and then edit it separately after successful delivery with HTML or Markdown* syntaxes.

I've verified that a new approach works for HTML, but it doesn't fully work for Markdown since Markdown is not currently working in the `master` branch version, either, and in Grafana Cloud, according to my tests.

Here is also some technical info for these changes:

## More

1. Webhook Infrastructure Enhancement (`http/client.go`, `http/webhook.go`, `receivers/webhook.go`)
- SendWebhook method signature changed: Now returns (`*receivers.WebhookResponse, error`) instead of just error
- Added WebhookResponse struct: Captures StatusCode, Body, and Headers from HTTP responses
- Enhanced error handling: Maintains response data even when validation fails
- Purpose: Enables receivers to access HTTP response details for processing (critical for Telegram message ID extraction)
2. Telegram Message Editing Feature (`receivers/telegram/v1/telegram.go`)
- Added message ID tracking: lastMessageID field with thread-safe access via mutex
- Two-step sending process:
   1. Send message with parse_mode="" (None) to ensure delivery
   2. Edit message with configured parse_mode to apply formatting
- New methods:
   - `EditMessage()`: Edits existing messages using Telegram's `editMessageText` API
   - `extractMessageID()`: Parses Telegram API response to extract message ID
   - `GetLastMessageID()` / `SetLastMessageID()`: Thread-safe message ID management
- API endpoint support: Added `editMessageText` alongside existing `sendMessage` and `sendPhoto`
3. Test Updates (`receivers/telegram/v1/telegram_test.go`)
- Updated expectations: Tests now expect two webhook calls per message (send + edit)
- Parse mode handling: First call sends without formatting, second applies configured parse mode
- Response mocking: Added mock response bodies with message IDs for testing
4. All Receivers Updated (19 receiver implementations)
- Signature compatibility: All SendWebhook calls updated to handle new return signature
- Minimal impact: Most receivers just ignore the response with _ since they don't need it
- Maintained functionality: No behavioral changes, just API compatibility


fopina:
I like the idea: always send none and let edit be the one to fail silently!

I can’t really comment on the code though couldn’t the change be restricted to telegram webhook only? So it would send and edit in the “send” method?
Avoiding impact in every webhook due to function signature changes

danilgotvyansky:
@fopina Yeah, I tried like that at first, but since the edit action requires response details, it would require either Telegram using an exclusive SendWebhook function or adding a response into an existing one, which requires all other receivers to expect a response and do nothing with it. 

The second approach looked easier :)

danilgotvyansky:
@soniaAguilarPeiron I see that you are assigned to the referenced issue. Could I ask you to take a look at this please?

PR #521, opened 2026-03-26
Set To header to undisclosed-recipients when using BCC

alexander-akhmetov:
When UseBCC is enabled, the To header was left empty, which looks odd to recipients and may cause some mail servers to flag the message as spam. Set it to `undisclosed-recipients:;` per [RFC 2822](https://datatracker.ietf.org/doc/html/rfc2822).

---

> [!NOTE]
> **Low Risk**
> Low risk: small header change limited to the `UseBCC && !SingleEmail` path, with tests updated to lock in the new `To`/`Bcc` behavior.
> 
> **Overview**
> When `EmailSenderConfig.UseBCC` is enabled (and `SingleEmail` is false), outgoing emails now set the `To` header to `undisclosed-recipients:;` while placing actual recipients in `Bcc`, instead of leaving `To` empty.
> 
> Tests in `email_sender_test.go` were updated to expect the new `To` header value in this scenario.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4da7d68235aae427117e24448e8955b275003cc0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/receivers/email_sender.go
+++ b/receivers/email_sender.go
@@ -236,6 +236,7 @@ func (s *defaultEmailSender) buildEmail(msg *Message) *gomail.Message {
 	}
 	m.SetHeader("From", msg.From)
 	if s.cfg.UseBCC && !msg.SingleEmail {
+		m.SetHeader("To", "undisclosed-recipients:;")
 		m.SetHeader("Bcc", msg.To...)
 	} else {
 		m.SetHeader("To", msg.To...)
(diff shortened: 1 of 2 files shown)


alexander-akhmetov:
bugbot review

aifraenkel:
If I may to chip-in, the PR looks good for a try. I would just evaluate having, or not a FT to rollback quick in case of unexpected issues.

alexander-akhmetov:
The tests break because `gomail.Send` doesn't like this address in To and fails

titolins (approved):

PR #538, opened 2026-04-27
Make logs for notification-related events consistent

santihernandezc:
### Problem

Notification events (successes, errors) are logged using arbitrary messages.  Some of the different possibilities to be aware of when searching through logs:
- Slack: `"Failed to send Slack message"`
- Google Chat: `"failed to send Google Hangouts Chat alert"`
- Line:  `"failed to send notification to LINE"`
- Threema: `"Failed to send threema notification"`
- Sensu Go: `"failed to send Sensu Go event"`
- MQTT: `"Failed to publish MQTT message"`
- Kafka: `"Failed to send notification to Kafka"`
- Victor Ops: `"failed to send notification"`
- SNS: `"Failed to publish to Amazon SNS. "` **(period + trailing space)**
- Dingding, Email, Jira, OnCall, OpsGenie, Pagerduty, Telegram, Webex, Webhook: No error

It's difficult to look for errors in integrations:
- Messages are arbitrary
- There's no consistency in how we use different terms ("event", "notification", "message" are used)
- Integration names can be capitalized, lowercase, uppercase, or not included in the message
- Events are logged only in some integrations
- The log level varies from `warn` to `error`

### Proposed solution

This PR makes logging consistent throughout all our integrations. Rules:
- The `msg` is always `"Failed to send notification"`
- Errors are logged with `warn` level (the caller logs with `error` level if all attempts fail)
- Successes are logged with `debug` level

Homogenizing how we log these events would allow operators to look for errors using a single query, filtering by integration. The pattern for searching error attempts becomes predictable. We can look for `msg="Failed to send notification"` and filter using the integration's name.

```logql
{job="<job>"} |= "Failed to send notification" |~ "slack|googlechat|..."
```

This also makes it easier to create graphs from these queries (e.g., using (`sum(rate(...))`).

These error/success logs also gives us visibility over receivers with multiple integrations. When a notify attempt fails for one of these, we can't know if any of the integrations succeeded in sending the notification and how many times.

Other improvements:
- `body` split over `request_body` and `response_body` in integrations logging either of them
- Error message and alert count added wherever missing

---

> [!NOTE]
> **Low Risk**
> Broad but mechanical changes across many receiver integrations to standardize log messages and fields; low functional risk but could affect operators relying on previous log text/levels.
> 
> **Overview**
> Adds centralized notification logging helpers (`receivers/logging.go`) with canonical `msg` values (**"Notification sent"** / **"Failed to send notification"**) and optional structured fields (e.g. `status_code`, `request_body`, `response_body`), plus unit tests.
> 
> Updates many receiver integrations to call `LogNotificationSent`/`LogNotificationFailed` (or the `Base` methods) on success/failure, replacing ad-hoc log lines; Discord/Teams additionally capture HTTP response status/body for richer failure logs, and several legacy (`v0mimir*`) notifiers now defer logging based on the returned error.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b224adf1fd5359f07dc438da521f4dd2639bc8b4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


yuri-tceretian (commented):
Not sure I understand the purpose. Each notifier is run by a stage that already emits the same messages for all types of integrations. 
I think it's ok when each notifier emits only what it makes sense in its partcular case.

yuri-tceretian (commented):
My recommendation would be to change the Notify method signature and insted rety, err return more detailed response, e.g. `Notify() (Result{status, additional metadata, Retryable bool}, error) . 
Where result contains the notification result including error while the second return argument error is non-nil in exceptional cases and automatically treated as non-retryable.

santihernandezc:
@yuri-tceretian This PR is only meant to change how we log events in our notifiers. I want to avoid changing the signature of any function. 

I updated the description to make the PR's intention clearer. It's still a WIP. I'm considering removing the helper functions and just modifying the message strings we have now to make them consistent.

PR #546, opened 2026-05-13
Include HTTP client response in contact point errors

santihernandezc:
If a webhook fails, we only show (e.g.) _"webhook response status 400"_ in the UI, no error message. This PR adds the response body (truncated) to the error we return.

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -22,7 +22,7 @@ import (
 	"github.com/grafana/alerting/receivers"
 )
 
-const maxResponseBodyInError = 1024
+const maxErrorResponseSize = 1024
 
 var ErrInvalidMethod = errors.New("webhook only supports HTTP methods PUT or POST")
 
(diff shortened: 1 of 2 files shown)


http/client.go
```diff
@@ -182,18 +184,22 @@ func (ns *Client) SendWebhook(ctx context.Context, l log.Logger, webhook *receiv
...
 	}
 
-	level.Debug(l).Log("msg", "Webhook failed", "url", url.Redacted(), "statuscode", resp.Status, "body", string(body))
-	return fmt.Errorf("webhook response status %v", resp.Status)
+	level.Debug(l).Log("msg", "Webhook failed", "url", url.Redacted(), "status_code", resp.Status, "body", string(body))
+	truncatedBody := string(body)
+	if len(truncatedBody) > maxResponseBodyInError {
+		truncatedBody = truncatedBody[:maxResponseBodyInError] + "... (truncated)"
+	}
+	return fmt.Errorf("webhook response status %v: %s", resp.Status, truncatedBody)
```
titolins: I wonder if we can't standardize this a bit and avoid reading the entire body in case of errors from the webhook. We could move the status check before reading the full body and use `io.LimitReader` instead - e.g. how it's done here: https://github.com/grafana/grafana/pull/124600/changes). WDYT?

santihernandezc: Some integrations still use `Validate` even when the status code is 2xx. We can't know beforehand if the request succeeded or not, so checking the status code is not enough.

yuri-tceretian (approved):
LGTM

PR #558, opened 2026-05-20
fix(telegram): accept numeric chatid and message_thread_id

### LINKED ISSUE #69950: Product-Area-Here: short description of bug here

yuri-tceretian:
Fixes parsing of Telegram's `chatid` and `message_thread_id` fields when provisioned via environment variables. Grafana evaluates `ENV="1"` as the integer `1` since grafana/grafana#63085, which caused:

```
json: cannot unmarshal number into Go struct field Config.chatid of type string
```

This ports the fix from #198 to the current versioned package structure (`receivers/telegram/v1`), using the existing `receivers.OptionalNumber` type — which already handles string-or-number JSON fields — rather than an ad-hoc `any` type switch.

Fixes grafana/grafana#69950

### DIFF
--- a/receivers/telegram/v1/config.go
+++ b/receivers/telegram/v1/config.go
@@ -32,32 +32,51 @@ type Config struct {
 }
 
 func NewConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Config, error) {
-	settings := Config{}
-	err := json.Unmarshal(jsonData, &settings)
-	if err != nil {
-		return settings, fmt.Errorf("failed to unmarshal settings: %w", err)
+	raw := struct {
+		BotToken              string                   `json:"bottoken,omitempty" yaml:"bottoken,omitempty"`
+		ChatID                receivers.OptionalNumber `json:"chatid,omitempty" yaml:"chatid,omitempty"`
+		MessageThreadID       receivers.OptionalNumber `json:"message_thread_id,omitempty" yaml:"message_thread_id,omitempty"`
+		Message               string                   `json:"message,omitempty" yaml:"message,omitempty"`
+		ParseMode             string                   `json:"parse_mode,omitempty" yaml:"parse_mode,omitempty"`
+		DisableWebPagePreview bool                     `json:"disable_web_page_preview,omitempty" yaml:"disable_web_page_preview,omitempty"`
+		ProtectContent        bool                     `json:"protect_content,omitempty" yaml:"protect_content,omitempty"`
+		DisableNotifications  bool                     `json:"disable_notifications,omitempty" yaml:"disable_notifications,omitempty"`
+	}{}
+	if err := json.Unmarshal(jsonData, &raw); err != nil {
+		return Config{}, fmt.Errorf("failed to unmarshal settings: %w", err)
 	}
-	settings.BotToken = decryptFn.Get("bottoken", settings.BotToken)
+
+	settings := Config{
+		Message:               raw.Message,
+		ParseMode:             raw.ParseMode,
+		DisableWebPagePreview: raw.DisableWebPagePreview,
+		ProtectContent:        raw.ProtectContent,
+		DisableNotifications:  raw.DisableNotifications,
+	}
+
+	settings.BotToken = decryptFn.Get("bottoken", raw.BotToken)
 	if settings.BotToken == "" {
 		return settings, errors.New("could not find Bot Token in settings")
 	}
+
+	settings.ChatID = raw.ChatID.String()
 	if settings.ChatID == "" {
 		return settings, errors.New("could not find Chat Id in settings")
 	}
+
 	if settings.Message == "" {
 		settings.Message = templates.DefaultMessageEmbed
 	}
 
-	var messageThreadID int
-	if settings.MessageThreadID != "" {
-		messageThreadID, err = strconv.Atoi(settings.MessageThreadID)
+	if raw.MessageThreadID != "" {
+		messageThreadID, err := strconv.Atoi(raw.MessageThreadID.String())
 		if err != nil {
 			return settings, errors.New("message thread id must be an integer")
 		}
-
 		if messageThreadID != int(int32(messageThreadID)) {
 			return settings, errors.New("message thread id must be an int32")
 		}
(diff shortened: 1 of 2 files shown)


receivers/telegram/v1/config.go
```diff
@@ -32,32 +32,51 @@ type Config struct {
...
 
 func NewConfig(jsonData json.RawMessage, decryptFn receivers.DecryptFunc) (Config, error) {
-	settings := Config{}
-	err := json.Unmarshal(jsonData, &settings)
-	if err != nil {
-		return settings, fmt.Errorf("failed to unmarshal settings: %w", err)
+	raw := struct {
+		BotToken              string                   `json:"bottoken,omitempty" yaml:"bottoken,omitempty"`
+		ChatID                receivers.OptionalNumber `json:"chatid,omitempty" yaml:"chatid,omitempty"`
+		MessageThreadID       receivers.OptionalNumber `json:"message_thread_id,omitempty" yaml:"message_thread_id,omitempty"`
```
titolins, line 38: Ran some tests on this with Claude and it turns out when these fields aren't pointers, null gets parsed as the literal string "null" and slips past the empty check. Making them pointers fixes it (Go's JSON decoder sets a pointer to nil for null instead of calling UnmarshalJSON), but we'd also need to check for nil below.

```suggestion
		ChatID                *receivers.OptionalNumber `json:"chatid,omitempty" yaml:"chatid,omitempty"`
		MessageThreadID       *receivers.OptionalNumber `json:"message_thread_id,omitempty" yaml:"message_thread_id,omitempty"`
```

receivers/telegram/v1/config.go
```diff
@@ -32,32 +32,51 @@ type Config struct {
...
 	}
+
 	if settings.Message == "" {
 		settings.Message = templates.DefaultMessageEmbed
 	}
 
-	var messageThreadID int
-	if settings.MessageThreadID != "" {
-		messageThreadID, err = strconv.Atoi(settings.MessageThreadID)
+	if raw.MessageThreadID != "" {
```
titolins, line 71:
```suggestion
	if raw.MessageThreadID != nil && raw.MessageThreadID.String() != "" {
```

receivers/telegram/v1/config.go
```diff
@@ -32,32 +32,51 @@ type Config struct {
...
+		ProtectContent:        raw.ProtectContent,
+		DisableNotifications:  raw.DisableNotifications,
+	}
+
+	settings.BotToken = decryptFn.Get("bottoken", raw.BotToken)
 	if settings.BotToken == "" {
 		return settings, errors.New("could not find Bot Token in settings")
 	}
+
+	settings.ChatID = raw.ChatID.String()
```
titolins, line 62:
```suggestion
	if raw.ChatID == nil {
		return settings, errors.New("could not find Chat Id in settings")
	}
	settings.ChatID = raw.ChatID.String()
```

receivers/telegram/v1/config.go
yuri-tceretian: Good point on the null case. Rather than switching to pointer fields, we are fixing this upstream in OptionalNumber.UnmarshalJSON — a null guard now decodes JSON null to an empty OptionalNumber instead of the literal string "null". This keeps all call sites clean with no nil checks needed. Fix is in the companion PR #559 (fix/optional-number-null).

receivers/telegram/v1/config.go
yuri-tceretian: Good point on the null case. Rather than switching to pointer fields, we are fixing this upstream in OptionalNumber.UnmarshalJSON — a null guard now decodes JSON null to an empty OptionalNumber instead of the literal string "null". This keeps all call sites clean with no nil checks needed. Fix is in the companion PR #559 (fix/optional-number-null).

receivers/telegram/v1/config.go
yuri-tceretian: Good point on the null case. Rather than switching to pointer fields, we are fixing this upstream in OptionalNumber.UnmarshalJSON — a null guard now decodes JSON null to an empty OptionalNumber instead of the literal string "null". This keeps all call sites clean with no nil checks needed. Fix is in the companion PR #559 (fix/optional-number-null).

titolins (approved):

PR #560, opened 2026-05-22
notify: remove MuteTimeIntervals from NotificationsConfiguration

yuri-tceretian:
## Summary

- Removes `MuteTimeIntervals []MuteTimeInterval` from `NotificationsConfiguration` — callers should place all intervals (mute and active) in `TimeIntervals`
- Removes `MuteTimeIntervals` from `ConfigFingerprint` and its hash computation (it was never being computed)
- Simplifies `buildTimeIntervals` to accept only `[]TimeInterval`
- Updates tests to use `TimeIntervals` for all interval entries

### DIFF
--- a/notify/grafana_alertmanager.go
+++ b/notify/grafana_alertmanager.go
@@ -152,12 +152,11 @@ type DynamicLimits struct {
 type NotifyReceiver = nfstatus.Receiver
 
 type NotificationsConfiguration struct {
-	RoutingTree       *Route
-	InhibitRules      []InhibitRule
-	MuteTimeIntervals []MuteTimeInterval
-	TimeIntervals     []TimeInterval
-	Templates         []templates.TemplateDefinition
-	Receivers         []models.ReceiverConfig
+	RoutingTree   *Route
+	InhibitRules  []InhibitRule
+	TimeIntervals []TimeInterval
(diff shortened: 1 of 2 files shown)


notify/grafana_alertmanager.go
```diff
@@ -210,7 +207,6 @@ func CalculateConfigFingerprint(cfg NotificationsConfiguration) ConfigFingerprin
 	fp.Overall = overallHash(
 		fp.RoutingTree,
 		fp.InhibitRules,
-		fp.MuteTimeIntervals,
```
moustafab, line 213: Doesn't this mean that it was being calculated as part of the hash previously? What's the impact of this hash changing?

moustafab (commented):
Question about the impact

notify/grafana_alertmanager.go
yuri-tceretian: The only impact is that the config will be re-applied, which will happen anyway on restart. In cloud it will just cause config update.

yuri-tceretian: Mute timing field is very old and usually not used. API always writes into time-intervals

JacobsonMT (approved):

PR #572, opened 2026-05-27
CI: Add TruffleHog secret scanning workflow

rwwiv:
Adds a TruffleHog secret scanning CI workflow that runs on PRs, pushes to main, and merge queue.

The workflow calls the reusable scanner from `grafana/security-github-actions`, which:
- Scans only changed files on PRs (full filesystem on push to main)
- Posts PR comments with findings and remediation instructions
- Applies org-wide exclude patterns (vendor dirs, lock files, etc.)
- Supports repo-local `.trufflehogignore` for additional exclusions
- Uploads scan artifacts for triage

Configuration:
- **Blocks on verified secrets** (`fail-on-verified: true`)
- **Does not block on unverified secrets** (`fail-on-unverified: false`)
- Sends scan metrics to Prometheus via Grafana Bench

### DIFF
--- a/.github/workflows/trufflehog.yml
+++ b/.github/workflows/trufflehog.yml
@@ -1,0 +1,26 @@
+name: TruffleHog Secret Scanning
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+  push:
+    branches:
+      - main
+  merge_group:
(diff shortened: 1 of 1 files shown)


yuri-tceretian (approved):
LGTM

PR #573, opened 2026-05-28
Remove *File, *Ref, and Files fields from v0 integration and HTTP config structs

yuri-tceretian:
## Summary

- Removes all `*File` fields from every v0mimir1/v0mimir2 receiver integration config struct (discord, email, slack, opsgenie, victorops, pagerduty, pushover, teams/v0mimir1, teams/v0mimir2, telegram, webhook)
- Removes all `*File`, `*Ref`, and `Files` fields from the shared `http/v0mimir` HTTP config structs (`BasicAuth`, `Authorization`, `OAuth2`, `HTTPClientConfig`, `TLSConfig`, `Header`)
- Simplifies `ValidateAlertmanagerConfig` in `mimir_validation.go`: per-integration file checks are no longer needed since the fields no longer exist in the structs
- Updates `definition/compat/receiver_compat.go` and `definition/compat.go` to drop the removed field assignments
- One commit per struct throughout

These fields reference the filesystem (`*File`) or a secret manager (`*Ref`), neither of which is available in the Grafana/Mimir alertmanager context. `ValidateAlertmanagerConfig` was already blocking them at runtime; this change enforces the restriction at the type level.


definition/mimir_validation.go
```diff
@@ -10,34 +10,17 @@ import (
 	commoncfg "github.com/prometheus/common/config"
```
yuri-tceretian, line 10: I will delete this file in a follow-up PR

titolins (approved):

PR #606, opened 2026-06-10
notify/historian: redact URLs in persisted notification errors

titolins:
Notification errors persisted to the history can contain the destination URL (e.g. transport failures surface the full request URL). This redacts URLs from the error string before it's written, reusing the existing `redactURL` helper.

### DIFF
--- a/http/client.go
+++ b/http/client.go
@@ -209,6 +209,16 @@ func redactURL(err error) error {
 	return e
 }
 
+// RedactError returns err as a string with any embedded request URL redacted,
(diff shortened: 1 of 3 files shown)


yuri-tceretian (dismissed):
LGTM but it took me a minute to understand tests. I wonder if it could be clearer

titolins:
> LGTM but it took me a minute to understand tests. I wonder if it could be clearer

yeah, that's a fair point. simplified now to comparing the full string 👍

yuri-tceretian (approved):

PR #615, opened 2026-06-29
Historian: Add plumbing for a minimal operator.

stevesg:
Add a standalone historian-operator command that runs the alerting historian
app via the grafana-app-sdk operator runner, and a bare bones unit test to
check it comes up and somewhat works. A more comprehensive integration test
would be preferable in the future.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ vendor
 .idea
 
 testing/alerting-gen/alerting-gen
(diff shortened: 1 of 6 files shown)


titolins (dismissed):
We should probably fix this: https://github.com/grafana/alerting/pull/615/changes#r3491710150
But could come separately also

titolins (approved):

PR #617, opened 2026-07-02
receivers: apply extra data to remaining v1 integrations

khalilhaji:
## What

Wires up `receivers.ApplyExtraData` in the remaining v1 receivers so that extra data provided via context is applied to the alert data before templating, matching the integrations that already support it:

`dingding`, `discord`, `googlechat`, `jira`, `kafka`, `line`, `pushover`, `sensugo`, `sns`, `teams`, `telegram`, `threema`, `victorops`, `wecom`.

## Tests

Adds a `TestNotify_ExtraData` for each receiver following the established pattern used by the already-covered integrations (opsgenie, webex, pagerduty, slack, etc.): the message template renders `$a.ExtraData`, the extra data is passed via context, and the rendered request body is asserted to contain it.

For **SNS**, its existing tests only exercised `createPublishInput` directly — `Notify` built a real AWS SNS client with no injectable mock. To test the `ApplyExtraData` call in `Notify`, this introduces a small `snsClient` interface and an overridable client factory field on the `Notifier`, so `Notify` can run against a mock client that captures the published message. `New`'s signature is unchanged.

All modified receiver packages' test suites pass; `go vet` and `gofmt` are clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/receivers/dingding/v1/dingding.go
+++ b/receivers/dingding/v1/dingding.go
@@ -40,7 +40,9 @@ func (dd *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error
 	dingDingURL := buildDingDingURL(dd.tmpl.ExternalURL, l)
 
 	var tmplErr error
-	tmpl, _ := templates.TmplText(ctx, dd.tmpl, as, l, &tmplErr)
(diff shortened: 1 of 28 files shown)


fayzal-g (commented):
lgtm, just extend for Kafka v3

fayzal-g (approved):

PR #622, opened 2026-07-14
Deps: Bump golang.org/x/crypto to v0.54.0

rwwiv:
Bumps `golang.org/x/crypto` to v0.54.0 (latest) to remediate reported security vulnerabilities.

The dependency is indirect (build-graph only; no package in this module imports it directly). Pinning it raises the MVS floor above every transitive requirement. `go mod tidy` also carried `golang.org/x/net` from v0.52.0 to v0.56.0 in the root module.

Applied to both the root module and `apps/historian`. Builds pass in both.


aespuna:
Done in https://github.com/grafana/alerting/pull/657

PR #628, opened 2026-07-21
RBAC for Notification History via the folder service

fayzal-g:
### Background
Notification History RBAC (added in #616) resolved a caller's accessible alert rules by listing `rules.alerting.grafana.app` through the Kubernetes rules API and filtering results to those rules (and the folders they live in). That approach does not work for the standalone historian. If the historian is deployed as a **multi-tenant** app, the rules API is still **single tenant**, so there is no way to point the historian at a per-tenant rules API - the configuration simply can't be expressed correctly in a multi-tenant deployment.

### What changed
RBAC is now resolved entirely through **multi-tenant** services:
1. **Enumerate folders** via the multi-tenant folder API (`folder.grafana.app`), forwarding the request identity.
2. **Authorize each folder** for `alert.rules:read` via an injected `authlib.AccessClient` (`BatchCheck`, folder-scoped, chunked by `MaxBatchCheckItems`). This exercises the alert-rules → folder mapping recently added to the authz service.
3. Restrict queries to the resulting accessible folder set using the existing LogQL folder push-down (`folder_uids` / `folder_uid` structured metadata), including the size-aware batching for large folder sets.

The `AccessClient` is supplied programmatically through `NotificationConfig` (not a flag), so `notification.New()` and `app.go` keep their signatures; the concrete client is wired by the deployment.

### Semantics change (folder-granular RBAC)
Alert-rule RBAC is strictly folder-scoped, so "can read a folder" ⇔ "can read every alert rule in it". Filtering at the folder level is therefore equivalent to per-rule RBAC **without enumerating individual rules** — which the single-tenant rules API can
no longer provide here. As a result, all per-rule post-processing is removed:
- No per-rule stripping of `RuleUIDs` from entries.
- No under-the-hood `rule_uids` grouping / fail-closed guard for counts.
- A notification admitted via an accessible folder is returned intact, including rule UIDs it co-references from other folders.
Alert lines remain filtered per folder (single-valued `folder_uid`), so per-alert access is unchanged. Multi-batch count results are merged via `mergeCounts` / `mergeRangeCounts`.

### Files
- `config.go` — add `AccessClient` to `NotificationConfig`; update RBAC docs; `authlib/types` promoted to a direct dependency.
- `rbac.go` — replace `k8sRuleAccessReader` with `k8sFolderAccessReader` (folder list + per-folder `alert.rules:read` check); `ruleFilter` is folder-only.
- `notification.go` — construct the folder reader; fail closed when RBAC is enabled but no `AccessClient` is configured.
- `lokireader.go` — remove all rule-level filtering; access enforced by the folder push-down alone.
- Tests updated to folder-only semantics.

### Testing
- `go build ./...`, `go vet ./...`, `go test ./pkg/app/notification/` pass; gofmt clean.
- The Loki integration test (`TestIntegration_NotificationRBAC`) self-skips unless
  `HISTORIAN_LOKI_URL` is set.

### DIFF
--- a/apps/historian/pkg/app/notification/rbac.go
+++ b/apps/historian/pkg/app/notification/rbac.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"strings"
 
+	authtypes "github.com/grafana/authlib/types"
 	"github.com/grafana/grafana-app-sdk/k8s"
 	"github.com/grafana/grafana-app-sdk/logging"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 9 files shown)


apps/historian/pkg/app/notification/rbac.go
```diff
@@ -205,86 +198,100 @@ func (s ruleUIDSet) sorted() []string {
...
 		}
 
 		raw, err := req.Do(ctx).Raw()
 		if err != nil {
-			return accessScope{}, fmt.Errorf("list alert rules: %w", err)
+			return nil, fmt.Errorf("list folders: %w", err)
 		}
 
-		var page partialRuleList
+		var page partialFolderList
```
titolins, line 287: One thing claude brought up is that by filtering by folders initially we might ignore users which `alert.rules:read` but not `folders:read`. Fixed roles bundle those together but there might be custom roles without it. I'm not sure this is possible, but I'd guess yes? Not sure there's much we can do about it though..

titolins (approved):

PR #643, opened 2026-07-27
Split legacy Mimir receiver config out of PostableApiReceiver

yuri-tceretian:
## Summary
- `PostableApiReceiver` no longer embeds the legacy Mimir `Receiver` struct — it's now Grafana-only (`Name` + `PostableGrafanaReceivers`).
- The legacy Mimir receiver struct moved to `definition/compat` as `compat.Receiver`.
- `LoadCompat` no longer applies per-integration Mimir defaults (Slack API URL, SMTP smarthost, etc.) since `PostableApiReceiver` can no longer carry that data.
- `PostableMimirReceiverToPostableGrafanaReceiver` now takes a `compat.Receiver` directly and returns a new `PostableApiReceiver` with converted Grafana-native integrations.


JacobsonMT (approved):
LGTM 🚀 Nice cleanup

PR #655, opened 2026-08-25
chore: remove the retired LINE integration

### LINKED ISSUE #334: Remove LINE integration
According to https://notify-bot.line.me/closing-announce
the integration EOL was in March, 2025. 
Therefore, it does not make sense to maintain the integration anymore. We should delete the support for integration.

We need to figure out how users upgrade Grafana, in case configuration contains LINE integrations.

loglapa:
## What this changes

- remove the retired LINE contact-point manifest and implementation
- retain a compatibility tombstone so existing persisted LINE configurations can still be decoded and skipped safely
- update receiver registration and tests to prevent LINE from being offered for new configurations

## Testing

- `go test ./notify/... ./receivers/...`
- `git diff --check`

Tests were run in Docker.

Fixes #334


loglapa:
Resolved the conflict with current main in b561a01. Main's only overlapping change was the generic-factory migration inside the LINE implementation; this PR intentionally removes that retired implementation, while retaining the LINE compatibility tombstone in the receiver type registry.

Validation passed: `git diff --check` and `go test ./notify/... ./receivers/...`.

PR #670, opened 2026-09-16
Fix discrepancies between config and schema for some integrations

yuri-tceretian:
Some v0 integrations have some incorrect schema fields. This PR fixes them and adds a test to keep them in sync

### DIFF
--- a/notify/schema_test.go
+++ b/notify/schema_test.go
@@ -80,7 +80,7 @@ func TestGetSecretKeysForContactPointType(t *testing.T) {
 		{receiverType: threema.Type, version: schema.V1, expectedSecretFields: []string{"api_secret"}},
 		{receiverType: opsgenie.Type, version: schema.V1, expectedSecretFields: []string{"apiKey"}},
 		{receiverType: webex.Type, version: schema.V1, expectedSecretFields: []string{"bot_token"}},
-		{receiverType: sns.Type, version: schema.V1, expectedSecretFields: []string{"sigv4.access_key", "sigv4.secret_key"}},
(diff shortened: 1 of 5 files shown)


notify/schema_invariants_test.go
```diff
@@ -0,0 +1,303 @@
+package notify
```
yuri-tceretian, line 1: I have a plan to make it more generic. I will address it in another PR

receivers/pushover/v1/config.go
```diff
@@ -164,6 +164,12 @@ var Schema = schema.NewIntegrationSchemaVersion(schema.IntegrationSchemaVersion{
 			SelectOptions: pushoverSoundOptions,
 			PropertyName:  "okSound",
 		},
+		{
+			Label:        "Upload image",
+			Description:  "Upload an image with the notification. Enabled by default.",
```
konstantinmv: If this is enabled by default, don't we have to teach FE about this before? I'm looking at [this](https://grafana.sourcegraph.com/r/github.com/grafana/grafana/-/blob/public/app/features/alerting/unified/components/receivers/form/GrafanaReceiverForm.tsx?L38) and it initialises empty, therefore the form later on coalesces this into false, and it arrives to the BE as false as well. 
I may be missing something though!

yuri-tceretian: Good finding. The schema does not have any "default value" . So, I decided to just omit "enabled by default". Schema is mostly UI related artifact

konstantinmv (approved):

PR #684, opened 2026-09-18
Refactor integration factories to use typed configs

yuri-tceretian:
## Summary
- Introduce a generic integration factory and migrate receiver factories to typed parsers and notifier builders.
- Expose config types through the factory registry and use them in schema invariant tests.
- Add `notify.GetFactoryForIntegration[T]()` for argument-free lookup by config type, with duplicate-type protection and tests.

## Testing
- `make lint`
- `make test` (before the typed lookup addition)
- `go test -race ./notify -count=1` (after the typed lookup addition)

Related to grafana/alerting-squad#1214.


notify/schema.go
```diff
@@ -57,9 +57,10 @@ var (
...
-	allSchemas     map[schema.IntegrationType]receivers.Manifest
-	aliasToType    map[schema.IntegrationType]schema.IntegrationType
-	initSchemaOnce sync.Once
+	allSchemas            map[schema.IntegrationType]receivers.Manifest
+	aliasToType           map[schema.IntegrationType]schema.IntegrationType
+	initSchemaOnce        sync.Once
+	factoriesByConfigType map[reflect.Type]receivers.VersionFactory
 )
 
 func initSchemas() {
```
JacobsonMT, line 66: I know this method of lazy init isn't new in this PR, but since we're adding more branches capable of panic, can we add an explicit unit test that just runs `initSchemas` and ensures it doesn't panic?

There might be existing tests that happen to hit the lazy init, but I'd like something dedicated that won't accidentally be removed.

notify/schema_invariants_test.go
```diff
@@ -78,89 +47,93 @@ var (
...
-// schemaCase pairs one version of an integration's schema with the Go Config struct that backs it.
-// The schema/factory registry (receivers.Manifest/IntegrationVersionFactory) only carries
-// json.RawMessage-based factory functions, not the concrete Config type, so there's no way to
-// derive this table from the registry - it has to be listed explicitly.
-type schemaCase struct {
-	integration schema.IntegrationType
-	version     schema.Version
-	fields      []schema.Field
-	configType  reflect.Type
+func TestIntegrationSchemasMatchConfigStructs(t *testing.T) {
```
JacobsonMT, line 50: Nice!

JacobsonMT (approved):
LGTM, one non-blocking comment on a desired test.

JacobsonMT (approved):

